PR #3, opened 2024-06-12
CCM-5179: Added additional style to header component

m-salaudeen:
## Type of changes

## Checklist

---

### DIFF
--- a/src/app/app.scss
+++ b/src/app/app.scss
@@ -1,1 +1,11 @@
-@import 'assets/nhsuk-8.1.1.min.css'
+@import 'assets/nhsuk-8.1.1.min.css';
+
+html {
+  font-size: 100%;
+  background-color: #fff;
+}
+
+body {
+  background: #fff;
+  overflow-x: hidden;
+}
--- a/src/app/app.scss
+++ b/src/app/app.scss
@@ -1,6 +1,1 @@
 @import 'assets/nhsuk-8.1.1.min.css';
-
-html {
-  font-size: 100%;
-  background-color: #f0f4f5;
-}
--- a/src/app/app.scss
+++ b/src/app/app.scss
@@ -2,10 +2,5 @@
 
 html {
   font-size: 100%;
-  background-color: #fff;
-}
-
-body {
-  background: #fff;
-  overflow-x: hidden;
+  background-color: #f0f4f5;
 }
--- a/src/app/content/content.ts
+++ b/src/app/content/content.ts
@@ -7,7 +7,7 @@ const headerComponent = {
 };
 
 const homePage = {
-  title: 'Notify - Web template management',
+  title: 'NHS Notify - Template Management',
   description: 'Template management',
 };
 
--- a/src/app/content/content.ts
+++ b/src/app/content/content.ts
@@ -1,12 +1,24 @@
 const headerComponent = {
   title: '',
-  description: '',
+  links: {
+    logIn: 'Log in',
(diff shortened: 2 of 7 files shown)


src/app/app.scss
```diff
@@ -1 +1,11 @@
...
+@import 'assets/nhsuk-8.1.1.min.css';
+
+html {
+  font-size: 100%;
+  background-color: #fff;
+}
+
+body {
+  background: #fff;
+  overflow-x: hidden;
```
bhansell1, line 10: Why do we have an overflow problem?

m-salaudeen: We don't, and this is to prevent us from having it. It's the standard in CSS

bhansell1: This would hide any potential problems in the future. I'd rather see the overflow than something that is potentially cut off.

I think we something does have an overflow, it would be better to fix in the component rather than globally? 

Just reading about this as a practice I'm seeing more "it's bad" than "good".

src/app/app.scss
```diff
@@ -1 +1,11 @@
-@import 'assets/nhsuk-8.1.1.min.css'
+@import 'assets/nhsuk-8.1.1.min.css';
+
+html {
+  font-size: 100%;
+  background-color: #fff;
+}
+
+body {
+  background: #fff;
```
bhansell1, line 9: We don't need this. The Default background colour is not white. It's `#f0f4f5` which is already applied when importing the css file.

src/app/app.scss
```diff
@@ -1 +1,11 @@
-@import 'assets/nhsuk-8.1.1.min.css'
+@import 'assets/nhsuk-8.1.1.min.css';
+
+html {
+  font-size: 100%;
+  background-color: #fff;
```
bhansell1: same here, I don't think we need to specify these colours and font size?

src/app/app.scss
m-salaudeen: I have removed the body section

src/app/utils/concatClassNames.ts
```diff
@@ -0,0 +1,12 @@
+type ClassName = string | string[] | undefined | false | null;
+
+export default function concatClassNames(...classNames: ClassName[]): string {
```
bhansell1, line 3: Any thoughts on using `classnames` package instead of our own?

It would save use from having to worry about stripping any issues with the name?

m-salaudeen: We are trying not to rely too much on external packages so I think it works best for us to implement some of these trivial utils than using external packages

src/app/app.scss
m-salaudeen: taken this off too

bhansell1 (dismissed):

chris-elliott-nhsd (dismissed):

src/app/content/content.ts
```diff
@@ -1,12 +1,24 @@
...
   title: '',
-  description: '',
+  links: {
+    logIn: 'Log in',
+    logOut: 'Log out',
+  },
+};
+
+const homePage = {
+  title: 'Notify - Web template management',
```
mark-r-bjss, line 10: Lets go with:
"NHS Notify - Template Management"

mark-r-bjss (dismissed):
Approved subject to tests coming later.

mark-r-bjss (approved):

chris-elliott-nhsd (approved):

PR #80, opened 2024-09-09
CCM-5522: Save NHS App template

bhansell1:
## Description

- Creates templateStorage datastore - Intended to store all the templates 
- Maps session object to a template object
- Saves template to database on TemplateSubmit

## Type of changes

## Checklist

---


src/domain/errors/coded.error.ts
```diff
@@ -0,0 +1,8 @@
+export class CodedError extends Error {
+  constructor(
+    message: string,
+    public readonly code: string
```
bhansell1, line 4: I'd like people's opinions on this. ATM I'm just passing in a string for the type of error it is. Should we look to use error codes such as `DBERR001` which then maps to a specific scenario?

src/components/forms/SubmitTemplate/server-action.ts
```diff
@@ -0,0 +1,51 @@
...
+      RedirectType.replace
+    );
+  } catch (error) {
+    logger.error('Failed to submit template', {
+      error,
+      sessionId,
+    });
+
+    // TODO: We need an internal server error page
+    throw error;
```
bhansell1, line 49: Maybe remove this? 

I've intentionally left out logging on the above function calls and moved it up a level to be more of a catch-all.

src/components/forms/SubmitTemplate/server-action.ts
```diff
@@ -0,0 +1,50 @@
...
+  const sessionId = String(formData.get('sessionId'));
+
+  const session = await getSession(sessionId);
+
+  if (!session) {
+    return redirect('/invalid-session', RedirectType.push);
+  }
+
+  try {
+    if (session.templateType === 'UNKNOWN') {
```
chris-elliott-nhsd, line 18: worth checking that this is a recognised template type rather than checking it isn't unknown? as a guard against rogue POST requests

src/components/forms/SubmitTemplate/server-action.ts
```diff
@@ -0,0 +1,50 @@
...
+      throw new Error('Unknown template type');
+    }
+
+    const templateDTO = templateFromSessionMapper(
+      session.templateType,
```
chris-elliott-nhsd, line 23: passing in a field from the session and then passing in the whole session looks odd

tsconfig.json
```diff
@@ -20,6 +20,7 @@
     ],
     "paths": {
       "@/amplify_outputs.json": ["./amplify_outputs.json"],
+      "amplify/*": ["./amplify/*"],
```
chris-elliott-nhsd, line 23: this is just mapping 'amplify/' to 'amplify/' which seems a bit unnecessary

src/components/forms/SubmitTemplate/server-action.ts
bhansell1: So this is handled in both `createTemplateFromSession` which throws an error if the templateType is unknown or not configured. 

Similarly with the zod validation. If the`type` is not configured within zod's union the validation will fail.

chris-elliott-nhsd (dismissed):

.eslintrc.json
```diff
@@ -67,7 +67,8 @@
     "react/no-danger": "off",
     "no-param-reassign": "off",
     "react/require-default-props": "off",
-    "no-useless-constructor": "off"
+    "no-useless-constructor": "off",
+    "sonarjs/no-small-switch": "off"
```
harrim91, line 71: Should this be disabled here? I guess you're just disabling it temporarily for `createTemplateFromSession`, where you'd expect the `switch` to grow over time. When the `switch` starts to grow, having this disabled here you'll probably miss it and forget to turn it back on. I'd just disabling it inline with a comment saying to re-enable it when the time comes. Or just get rid of the switch for now!

src/components/forms/SubmitTemplate/server-action.ts
```diff
@@ -15,19 +15,9 @@ export async function submitTemplate(formData: FormData) {
...
-    const templateDTO = templateFromSessionMapper(
-      session.templateType,
-      session
-    );
-
-    const validatedTemplate = validateTemplate(
-      session.templateType,
-      templateDTO
-    );
+    const validatedTemplate = validateTemplate(templateDTO);
```
harrim91, line 20: Thinking a little about this, I feel the most pressing place that needs validation is shifting it left into `getSession`. There it's getting the session from the database, but it's getting it's type from an assertion (`return data as Session`), not from any actual validation. 

If the session data you get back from the database isn't valid, you're more liable to see a runtime error when calling `createTemplateFromSession` before you get to this validation. 

And then once you add the validation to the session, it seems like the `validateTemplate` function is kind of redundant, because you've already validated the session, and the template creation logic is entirely within the scope of this app. i.e. Given a valid session, should `createTemplateFromSession` ever return an invalid template?

harrim91 (commented):
some thoughts

.eslintrc.json
bhansell1: I don't think small switch statements are an issue, which Is why i disabled the rule entirely. The alternative is to use if/else based on the suggestion in https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-small-switch.md.

src/domain/templates/templates.types.ts
```diff
@@ -4,17 +4,21 @@ const baseFields = z.object({
...
+const $NHSAppTemplateSchema = $BaseTemplateSchema.extend({
   type: z.literal('NHS_APP'),
 });
 
-export type Template = z.infer<typeof $Template>;
+export const $TemplateSchema = z.discriminatedUnion('type', [
+  $NHSAppTemplateSchema,
+]);
+
+export type Template = z.infer<typeof $BaseTemplateSchema>;
```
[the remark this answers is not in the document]
bhansell1: The typing for NHSAppTemplate is a remnant of something I was playing around with. Good shout on the typing for Template!

src/components/forms/SubmitTemplate/server-action.ts
bhansell1: Hm... possibly? I like the fail fast/early and generally agree with you. I think my hesitancy is that the session object can be in different `valid`states depending on the `templateType` you choose. Accounting for each different valid state seems strange? What do you think? maybe another call?

src/app/nhs-app-template-submitted/[templateId]/page.tsx
```diff
@@ -1,8 +1,13 @@
...
 
-const NhsAppTemplateSubmittedPage = async () => (
+const NhsAppTemplateSubmittedPage = async ({
+  params,
+}: {
+  params: { templateId: string };
+}) => (
   <h2 className='nhsuk-heading-l' data-testid='page-sub-heading'>
     Placeholder page
+    <p>{params.templateId}</p>
```
Comment (mark-r-bjss) [inline src/app/nhs-app-template-submitted/[templateId]/page.tsx:10]: As per previous discussion, lets not put unsanitized user data directly into the page.

src/components/forms/SubmitTemplate/server-action.ts
```diff
@@ -0,0 +1,40 @@
+'use server';
+
+import { redirect, RedirectType } from 'next/navigation';
+import { getSession, saveTemplate } from '@utils/form-actions';
+import { createTemplateFromSession, validateTemplate } from '@domain/templates';
+import { logger } from '@utils/logger';
+
+export async function submitTemplate(formData: FormData) {
+  const sessionId = String(formData.get('sessionId'));
```
mark-r-bjss, line 9: The other server actions get this data a little differently, using `formData.get(...` and `schema.safeParse(...`

e.g.
src/components/forms/CreateNhsAppTemplate/server-action.ts#L52

src/components/forms/SubmitTemplate/server-action.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const session = await getSession(sessionId);
+
+  if (!session) {
+    return redirect('/invalid-session', RedirectType.push);
```
mark-r-bjss, line 14: Should be replace because we don't want to leave the bad URL in the history.  Also, I don't think you need to return, I think this throws (but please double check).

src/components/forms/SubmitTemplate/server-action.ts
bhansell1: We don't need a return but it helps with unit tests. If we don't return we end up having to mock out the functionality to cause the code to exit early - which is implemented in `__tests__/app/create-nhs-app-template/page.test.tsx`.

I think returning makes the unit tests a bit easier. But happy to follow the established pattern.

src/components/forms/SubmitTemplate/server-action.ts
bhansell1: updated

bhansell1: Also updated to replace.

src/app/nhs-app-template-submitted/[templateId]/page.tsx
Comment (bhansell1) [reply src/app/nhs-app-template-submitted/[templateId]/page.tsx]: updated. Good shout, thanks.

src/components/forms/SubmitTemplate/server-action.ts
bhansell1: Honestly think I'm faffing a bit too much about this. I'm going to leave it as for now and write up some tickets.

src/domain/templates/templates.types.ts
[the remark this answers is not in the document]
bhansell1: updated.

mark-r-bjss (dismissed):

chris-elliott-nhsd (dismissed):

harrim91 (approved):

PR #84, opened 2024-09-11
CCM-6319: initial framework, pipeline job and tests

sean-corrigan1:
## Description

Create initial framework for running automated UI component tests:
- Set up playwright dependencies
- Create initial config to allow local running of tests
- Create setup / teardown of session data to allow seeding of data in Amplifly dynamoDB
- Create Page Object Models for initial components (pages)
- Create initial tests for components (pages)
- Integrate UI component tests into CI/CD Github workflow
- Output test report as artefact with screenshots (on failure) from CI/CD Github workflow

## Context

To automate UI testing and minimise manual test effor

## Type of changes

## Checklist

- [x] This PR is a result of pair or mob programming

---

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -4,6 +4,7 @@ pre-commit 3.6.0
 terraform 1.9.2
 vale 3.6.0
 tfsec 1.28.10
+nodejs 20.13.1
 
 # ==============================================================================
 # The section below is reserved for Docker image versions.
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
     "ts-node": "^10.9.2",
     "tsx": "^4.19.1",
     "typescript": "^5.5.4",
-    "wait-on": "^8.0.0"
+    "wait-on": "^8.0.0",
+    "nhs-notify-web-template-management-ui-tests": "file:tests/test-team"
   },
   "overrides": {
     "pa11y-ci": {
--- a/tests/test-team/README.md
+++ b/tests/test-team/README.md
@@ -1,0 +1,8 @@
+# NHS Notify Template Management (WebUI) Automated Regression Tests
+
+This package includes tests for the template management UI. The tests can be run using the test:local-ui script in the package.json.
+
+This package includes:
+
+- component tests, which are designed to test the behaviour on individual pages, and seed session data relevant to those pages at the start of the tests
+- e2e tests, which simulate a user going through the whole app.
--- a/tests/test-team/config/playwright.config.ts
+++ b/tests/test-team/config/playwright.config.ts
@@ -1,0 +1,27 @@
+import { defineConfig } from '@playwright/test';
+
+export default defineConfig({
+  testDir: '../',
+  fullyParallel: false,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  retries: 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 4 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: [
+    ['line'],
+    [
+      'html',
+      {
+        outputFolder: '../playwright-report',
+        open: process.env.CI ? 'never' : 'on-failure',
+      },
+    ],
+  ],
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
+  },
+});
--- a/tests/test-team/package.json
+++ b/tests/test-team/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "nhs-notify-web-template-management-ui-tests",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "test:start-local-app": "npm run build --prefix ../../ && npm run start --prefix ../../",
+    "test:local-ui": "playwright test --project component -c config/local.config.ts",
+    "test:local-ui-e2e": "playwright test --project e2e-local -c config/local.config.ts"
+  },
+  "devDependencies": {
+    "@aws-sdk/client-appsync": "^3.650.0",
+    "@aws-sdk/client-dynamodb": "^3.654.0",
+    "@aws-sdk/lib-dynamodb": "^3.654.0",
+    "@playwright/test": "^1.45.1",
+    "aws-amplify": "^6.6.0"
+  }
+}
--- a/tests/test-team/pages/template-mgmt-base-page.ts
+++ b/tests/test-team/pages/template-mgmt-base-page.ts
@@ -1,0 +1,51 @@
+import { Locator, type Page } from '@playwright/test';
+
+export class TemplateMgmtBasePage {
+  readonly page: Page;
+
+  readonly notifyBannerLink: Locator;
+
+  readonly loginLink: Locator;
+
+  readonly goBackLink: Locator;
+
+  readonly pageHeader: Locator;
+
+  readonly continueButton: Locator;
+
+  constructor(page: Page) {
+    this.page = page;
+    this.notifyBannerLink = page.locator(
+      '[class="nhsuk-header__link nhsuk-header__link--service"]'
+    );
+    this.loginLink = page
+      .locator('[class="nhsuk-account__login--link"]')
+      .and(page.getByText('Log in'));
+    this.goBackLink = page
+      .locator('[class="nhsuk-back-link__link"]')
+      .and(page.getByText('Go back'));
+    this.pageHeader = page
+      .locator('[class="nhsuk-heading-xl"]')
+      .and(page.locator('h1'));
+    this.continueButton = page
+      .locator('[class="nhsuk-button"]')
+      .and(page.getByRole('button'))
+      .and(page.getByText('Continue'));
+  }
+
+  async navigateTo(url: string) {
+    await this.page.goto(url, { waitUntil: 'load' });
+  }
+
+  async clickNotifyBannerLink() {
+    await this.notifyBannerLink.click();
+  }
+
+  async clickLoginLink() {
+    await this.loginLink.click();
+  }
+
+  async clickContinueButton() {
+    await this.continueButton.click();
+  }
+}
--- a/tests/test-team/pages/template-mgmt-choose-page.ts
+++ b/tests/test-team/pages/template-mgmt-choose-page.ts
@@ -1,0 +1,33 @@
+import { Locator, type Page } from '@playwright/test';
+import { TemplateMgmtBasePage } from './template-mgmt-base-page';
+
+export class TemplateMgmtChoosePage extends TemplateMgmtBasePage {
+  readonly chooseTemplateRadioGroup: Locator;
+
+  readonly fieldsetHeading: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.chooseTemplateRadioGroup = page.getByRole('radiogroup');
+    this.fieldsetHeading = page.locator('[class="nhsuk-fieldset__heading"]');
+  }
+
+  static readonly templateOptions = [
+    'NHS App message',
+    'Email',
+    'Text message (SMS)',
+    'Letter',
+  ];
+
+  private static readonly chooseTemplatePageUrl = `/templates/choose-a-template-type`;
+
+  async navigateToChooseTemplatePage(sessionId: string) {
+    await this.navigateTo(
+      `${TemplateMgmtChoosePage.chooseTemplatePageUrl}/${sessionId}`
+    );
+  }
+
+  static async checkRadioButton(templateRadioButton: Locator) {
+    await templateRadioButton.check();
+  }
+}
--- a/tests/test-team/pages/template-mgmt-create-page.ts
+++ b/tests/test-team/pages/template-mgmt-create-page.ts
@@ -1,0 +1,11 @@
+import { TemplateMgmtBasePage } from './template-mgmt-base-page';
+
+export class TemplateMgmtCreatePage extends TemplateMgmtBasePage {
+  private static createTemplatePageUrl = '/templates/create-nhs-app-template';
+
+  async navigateToCreateNhsAppTemplatePage(sessionId: string) {
+    await this.navigateTo(
+      `${TemplateMgmtCreatePage.createTemplatePageUrl}/${sessionId}`
+    );
+  }
+}
--- a/tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
+++ b/tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
@@ -1,0 +1,171 @@
+import { test, expect } from '@playwright/test';
+import { TemplateMgmtChoosePage } from '../pages/template-mgmt-choose-page';
+import { Session, TemplateType } from '../helpers/types';
+import SessionStorageHelper from '../helpers/session-storage-helper';
+
+export const emptySessionData: Session = {
+  __typename: 'SessionStorage',
+  id: '3d98b0c4-6666-0000-0000-95eb27590000',
+  createdAt: '2024-09-19T23:36:20.815Z',
+  updatedAt: '2024-09-19T23:36:20.815Z',
+  templateType: 'UNKNOWN',
+  nhsAppTemplateName: '',
+  nhsAppTemplateMessage: '',
+};
+
+export const emptySessionDataForRadioSelect: Session = {
+  __typename: 'SessionStorage',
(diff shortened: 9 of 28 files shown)


package.json
```diff
@@ -10,6 +10,8 @@
     "lint:fix": "next lint  --dir . --fix",
     "test:unit": "jest",
     "test:accessibility": "pa11y-ci -c ./tests/accessibility/.pa11y-ci.js",
+    "test:local-ui": "npx playwright test --project component -c tests/test-team/config/local.config.ts",
```
mark-r-bjss, line 13: Should these be their own project/package.json - given that you can see some application errors during tests?

tests/test-team/pages/template-mgmt-choose-page.ts
```diff
@@ -0,0 +1,61 @@
...
+  constructor(page: Page) {
+    super(page);
+    this.chooseTemplateRadioGroup = page.getByRole('radiogroup');
+    this.fieldsetHeading = page.locator('[class="nhsuk-fieldset__heading"]');
+  }
+
+  static emptySessionData: Session = {
+    id: '3d98b0c4-6666-0000-0000-95eb27590000',
+    templateType: 'UNKNOWN',
+    nhsAppTemplateName: ' ',
```
mark-r-bjss, line 19: Shouldn't this be an empty string rather than a space?

tests/test-team/pages/template-mgmt-create-page.ts
```diff
@@ -0,0 +1,28 @@
+// import { type Page } from '@playwright/test';
```
mark-r-bjss, line 1: Can you check that the linter and typechecker are covering these files.  The linter should have picked up on these comments.

tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
```diff
@@ -0,0 +1,135 @@
...
+  test('should display NHS App radio button selected if present in session storage', async ({
+    page,
+  }) => {
+    const chooseTemplatePage = new TemplateMgmtChoosePage(page);
+
+    await chooseTemplatePage.navigateToChooseTemplatePage(
+      TemplateMgmtChoosePage.nhsAppRadioSelectedSessionData.id
+    );
+
+    // expect to assert radio selected
```
mark-r-bjss, line 123: Is this deliberate?

tests/test-team/setup/session.setup.ts
```diff
@@ -0,0 +1,29 @@
...
+    ...TemplateMgmtChoosePage.sessionData,
+    ...TemplateMgmtCreatePage.sessionData,
+  ];
+
+  Amplify.configure(config);
+
+  const client = generateClient<Schema>({ authMode: 'iam' });
+
+  for (const data of sessionData) {
+    await client.models.SessionStorage.create(data);
```
mark-r-bjss, line 27: The benefit of this is that it is quick to develop now but the downside is that the test is not independent of the application code because it is using the graphQL API to create the data.  An alternative would be to go directly to the database via an independent auth role.  I'm not sure how this approach will work when we have a more mature auth layer as we won't have guest access to the APIs.

tests/test-team/pages/template-mgmt-base-page.ts
```diff
@@ -0,0 +1,51 @@
...
+  readonly goBackLink: Locator;
+
+  readonly pageHeader: Locator;
+
+  readonly continueButton: Locator;
+
+  constructor(page: Page) {
+    this.page = page;
+    this.notifyBannerLink = page.locator(
+      '[class="nhsuk-header__link nhsuk-header__link--service"]'
```
bhansell1, line 19: What do you think about using a `data-*` attribute on HTML elements on the code instead of relying on `classnames` and `text` which are subject to change and in the case of a class return more than 1 object.

For example if we have some html:

```
<button data-test-id="pw-important-buttom">
Click me!
</button>
```

Your locators can use the data-* attribute like:

```
page.locator('data-testid=important-button')
```

I found this [article](https://playwrightsolutions.com/getbytestid/) which seems reasonably interesting and it looks like Playwright [has native support for finding elements by data-* attribute](https://playwrightsolutions.com/getbytestid/).

tests/test-team/setup/session.setup.ts
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable no-await-in-loop */
+/* eslint-disable unicorn/prefer-module */
+/* eslint-disable import/no-unresolved */
+/* eslint-disable @typescript-eslint/no-var-requires */
```
bhansell1, line 4: is this needed?

I generally follow the rule of them where I don't inline disable a rule, unless it's a super edge case. If the rule isn't valid or I disagree with it. Then I'd disable it globally or update the code to fix the linting issue. 

I think in the case of the ones you've disabled they might not be relevant any more? Apart from the `no-await-in-loop`.

Which is valid. You can update your code to promisify it like so;

```
const promises = sessionData.map((session) => client.create(session));

await Promise.all(promises);
```

This also has additional benefits of being slightly faster because it's "concurrent". Also if any of the promises fail then it'll throw an error. The alternative is `Promise.allSettled` which doesn't fail but you have to reason over the response.

tests/test-team/teardown/session.teardown.ts
```diff
@@ -0,0 +1,29 @@
...
+  const sessionData: Session[] = [
+    ...TemplateMgmtChoosePage.sessionData,
+    ...TemplateMgmtCreatePage.sessionData,
+  ];
+
+  Amplify.configure(config);
+
+  const client = generateClient<Schema>({ authMode: 'iam' });
+
+  for await (const data of sessionData) {
```
bhansell1, line 26: Same here you can Promisify it. Using Promise.all() or Promise.allSettled.

tests/test-team/pages/template-mgmt-choose-page.ts
sean-corrigan1: Yes ... sorted.

tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
sean-corrigan1: Yes, will be picked up by CCM-6321 to complete.

tests/test-team/setup/session.setup.ts
sean-corrigan1: Updated so as not to use the GraphQL api to create data ... the dynamo table name is ascertained by matching GraphQL api endpoints between that in the amplify outputs and that listed using the a method from AppSync SDK. Then uploaded directly to dynamo.

tests/test-team/pages/template-mgmt-base-page.ts
sean-corrigan1: This was a choice not to use the test id based on what i'm used to, and in my head, the html should be testable without the addition of special test attributes (yes, i disagree with an established test thing ... probably a losing battle). 

The fact they are subject to change makes it more valuable rather than an introduction of flakiness. While we are explicitly testing the functionality of, say a button, the use of class / role and text implicitly tests the button has the correct text and that the button is a button and not a link (both would would pass a functional test as they utilise the .click() method if using a data-testid locator that hasn't been updated).

If the element cant be found then something has gone wrong / changed by accident / unexpectedly, test id may not pick up any of this.

Happy to iterate on it / discuss if my logic sounds to picky / trying to do too much.

tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
```diff
@@ -0,0 +1,135 @@
...
+      'Choose a template type to create'
+    );
+
+    await expect(page.locator('[class="nhsuk-error-summary"]')).toBeVisible();
+
+    await expect(
+      page
+        .locator('[class="nhsuk-list nhsuk-error-summary__list"]')
+        .getByRole('listitem')
+    ).toHaveText(['Select a template type']);
```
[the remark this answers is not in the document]
sean-corrigan1: Something to think about, i may just move this to https://nhsd-jira.digital.nhs.uk/browse/CCM-6321 where we will review the current tests 👍 It is something thats dynamic on the page and i believe it modifies the url so maybe something to add in with above ticket and use it as a pattern moving forward.

tests/test-team/setup/session.setup.ts
sean-corrigan1: This is cool, got a bit stuck with the async/await stuff but this is nice and clean 🦖

package.json
sean-corrigan1: Updated to separate out package.json and removed npx bit

tests/test-team/pages/template-mgmt-create-page.ts
sean-corrigan1: Outdated, commented out code uncommented.

tests/test-team/pages/template-mgmt-base-page.ts
bhansell1: Hey, that's a completely valid reason! 

You're right in saying searching by specifics like by `text` or `role` is a good way of finding elements. 

My one concern is using class names in locators as that can be quite wide ranging. More of a nitpick than anything!

tests/test-team/template-mgmt-component-tests/template-mgmt-start-page.component.ts
```diff
@@ -83,6 +83,33 @@ test.describe('Start Page', () => {
...
+      name: 'Privacy Policy',
+      selector: 'a[data-testid="privacy-policy-link"]',
+    },
+    {
+      name: 'Terms and Conditions',
+      selector: 'a[data-testid="terms-and-conditions-link"]',
+    },
+  ];
+
+  Promise.all(
```
bhansell1, line 107: Probably need an await here.

tests/test-team/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "nhs-notify-web-template-management-ui-tests",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "test:start-local-app": "npm run dev --prefix ../../",
```
bhansell1, line 6: This is a really great idea!

I wonder whether we should be running against a production build vs running in dev. OFC running `npm run dev` is faster but doing a `npm run build && npm run start` is more closer aligned with the deployed version of the app. 

There are small differences with how the app works in both modes (I remember some time ago we had really flakey PW tests due to running in dev mode). 

I'll add a comment to one of the tickets you've raised.

tests/test-team/setup/session.setup.ts
```diff
@@ -0,0 +1,14 @@
+import { test as setup } from '@playwright/test';
+import { TemplateMgmtChoosePage } from '../pages/template-mgmt-choose-page';
+import { TemplateMgmtCreatePage } from '../pages/template-mgmt-create-page';
+import { Session } from '../helpers/types';
+import SessionStorageHelper from '../helpers/session-storage-helper';
+
+setup('setup session data', async () => {
+  const sessionData: Session[] = [
+    ...TemplateMgmtChoosePage.sessionData,
```
bhansell1, line 9: This seems a little strange having session data on the `TemplateMgmtChoosePage` POM. 

Is it worth splitting out the session data into a session fixture? 

I'll add the suggestion to your ticket.

tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
```diff
@@ -0,0 +1,135 @@
...
+    baseURL,
+  }) => {
+    const chooseTemplatePage = new TemplateMgmtChoosePage(page);
+
+    await chooseTemplatePage.navigateToChooseTemplatePage(
+      TemplateMgmtChoosePage.emptySessionData.id
+    );
+
+    await expect(page).toHaveURL(
+      `${baseURL}/templates/choose-a-template-type/3d98b0c4-6666-0000-0000-95eb27590000`
```
bhansell1, line 16: This is great!

I wonder if we could use the Id directly from the `sessionWithNoData` object so that it's a bit easier to tie this Id back to the scenario? 

Otherwise it just looks like a magic number (uuid) with no meaning.

package.json
```diff
@@ -24,6 +24,7 @@
     "markdown-it": "^13.0.1",
     "mimetext": "^3.0.24",
     "next": "14.2.6",
+    "nhs-notify-web-template-management-ui-tests": "file:tests/test-team",
```
chris-elliott-nhsd, line 27: what's the benefit of having this here? it seems like it will install all the test-related stuff whenever we do an npm install, including for production builds, which i don't think is ideal

tests/test-team/README.md
```diff
@@ -0,0 +1 @@
+# NHS Notify Template Management (WebUI) Automated Regression Tests
```
chris-elliott-nhsd, line 1: this empty README doesn't seem very useful - either add to it or remove it

tests/test-team/config/playwright.config.ts
```diff
@@ -0,0 +1,29 @@
+import { defineConfig } from '@playwright/test';
+
+export default defineConfig({
+  testDir: '../',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
```
chris-elliott-nhsd, line 10: do we want to allow retries in CI? i know we do this in Notify core but would it be good to start off here committing to writing non-flaky tests?

tests/test-team/README.md
bhansell1: Might be a good place to talk about the difference between the e2e tests and component tests and where the boundary is between them?

tests/test-team/pages/template-mgmt-base-page.ts
chris-elliott-nhsd: i agree with Sean here - i'm also not a fan of adding stuff to production code that's only for use by the tests

tests/test-team/config/playwright.config.ts
chris-elliott-nhsd: updated

tests/test-team/README.md
chris-elliott-nhsd: added

tests/test-team/setup/session.setup.ts
chris-elliott-nhsd: updated - test suites now seed their own data, none of this is global data so we don't really need a global setup at the moment

tests/test-team/package.json
chris-elliott-nhsd: updated

tests/test-team/template-mgmt-component-tests/template-mgmt-start-page.component.ts
chris-elliott-nhsd: updated

bhansell1 (dismissed):

package.json
```diff
@@ -20,7 +20,12 @@
   "dependencies": {
     "@aws-amplify/adapter-nextjs": "^1.2.19",
     "@aws-amplify/backend": "^1.2.0",
+    "@aws-sdk/client-appsync": "^3.650.0",
+    "@aws-sdk/client-dynamodb": "^3.654.0",
```
mark-r-bjss, line 24: The tests should operate as their own app with a separate package.json so that we don't pollute the production code with unnecessary test code or in this case dependencies.

chris-elliott-nhsd: updated

.tool-versions
```diff
@@ -4,6 +4,7 @@ pre-commit 3.6.0
 terraform 1.9.2
 vale 3.6.0
 tfsec 1.28.10
+nodejs 20.13.1
```
mark-r-bjss, line 7: Do we also need to make sure that the backend lambda runtimes stays in sync with this?

package.json
```diff
@@ -24,6 +24,7 @@
     "markdown-it": "^13.0.1",
     "mimetext": "^3.0.24",
     "next": "14.2.13",
+    "nhs-notify-web-template-management-ui-tests": "file:tests/test-team",
```
mark-r-bjss, line 27: Why do we need this?

.tool-versions
chris-elliott-nhsd: updated

package.json
chris-elliott-nhsd: moved to dev dependency as discussed

mark-r-bjss (approved):

mark-r-bjss (approved):

jazrattanpal (approved):
Approved by Test

bhansell1 (dismissed):

chris-elliott-nhsd (approved):

jazrattanpal (approved):

jazrattanpal (approved):

mark-r-bjss (approved):

PR #139, opened 2024-10-17
CCM-7283: Use self-container Cognito User Pool as a temporary auth solution

chris-elliott-nhsd:
## Description

This is a PR to use the user pool that is created as part of the Amplify backend for auth. This is a strictly temporary solution as we would like to move in the medium term to using the separate auth microservice, however this cannot be done until we have done some additional work to move away from Amplify data which does not support external auth providers.

## Type of changes

- [x] Breaking change (fix or feature that would change existing functionality)

## Checklist

---


tests/test-team/helpers/test-user-client.ts
```diff
@@ -0,0 +1,67 @@
...
+  private readonly userPoolId: string;
+
+  constructor(amplifyOutputsPathPrefix = '../..') {
+    this.userPoolId = new AmplifyAppSyncClient(
+      amplifyOutputsPathPrefix
+    ).getUserPoolId();
+  }
+
+  async createTestUser(email: string, password: string) {
+    const res = await this.cognitoClient.send(
```
mark-r-bjss, line 23: Do we want to check if it exists first or delete it just in case?

.gitleaksignore
```diff
@@ -2,3 +2,4 @@
 
 cd9c0efec38c5d63053dd865e5d4e207c0760d91:docs/guides/Perform_static_analysis.md:generic-api-key:37
 87312c6a627a7b0420956d49187fd15b130df170:src/__tests__/components/molecules/LoginStatus.test.tsx:jwt:23
+37ca9f5670f4cd7d91869845ca27defbe6156bb9:src/__tests__/components/molecules/LoginStatus.test.tsx:jwt:23
```
mark-r-bjss, line 5: Do we actually need this, the line only has `undefined` data?  Can we clear all these down?

tests/accessibility/create-test-user.ts
```diff
@@ -0,0 +1,4 @@
+import { TestUserClient } from 'nhs-notify-web-template-management-ui-tests/helpers/test-user-client';
+
+const testUserClient = new TestUserClient('.');
+testUserClient.createTestUser('accessibility-test@nhs.net', 'Test-Password1');
```
mark-r-bjss, line 4: Can we use a secure randomly generated password rather than hard coded to keep any deployed resources protected.

src/__tests__/components/forms/CreateSmsTemplate/CreateSmsTemplate.test.tsx
```diff
@@ -110,7 +110,7 @@ describe('CreateSmsTemplate component', () => {
       throw new Error('Template name box not found');
     }
 
-    const longMessage = 'x'.repeat(300);
+    const longMessage = 'x'.repeat(30);
```
mark-r-bjss, line 113: 30 chars is not a particularly long message, was this a deliberate change?

src/components/molecules/Header/Header.tsx
```diff
@@ -51,11 +54,7 @@ export function NHSNotifyHeader({ className, dataTestId }: HeaderType) {
           id='content-header'
         >
           {/* I am currently testing the link wrapper, this will change later when we implement auth as the link will change based on auth state */}
-          <div className='nhsuk-account__login' data-testid='login-link'>
-            <Link className='nhsuk-account__login--link' href='/'>
-              {content.components.headerComponent.links.logIn}
-            </Link>
-          </div>
+          {pathname?.includes('auth') ? '' : <LoginStatus />}
```
mark-r-bjss, line 57: Can we at least have a const somewhere for this?

src/components/molecules/LoginStatus/LoginStatus.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { authStatus } = useAuthenticator();
+  const [loggedInUser, setLoggedInUser] = useState<string>();
+
+  useEffect(() => {
+    fetchAuthSession().then((session) => {
+      setLoggedInUser(session.tokens?.idToken?.payload?.email?.toString());
+    });
+  }, [authStatus]);
+
+  if (authStatus === 'authenticated') {
```
mark-r-bjss, line 21: Is there an enum to reference?

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-require-imports,unicorn/prefer-module */
+
+import { Authenticator, useAuthenticator } from '@aws-amplify/ui-react';
+import { usePathname, redirect } from 'next/navigation';
+import { ReactNode, FC } from 'react';
+import { Amplify } from 'aws-amplify';
+import { getAmplifyOutputs } from '@utils/get-amplify-outputs';
+
+Amplify.configure(getAmplifyOutputs(), { ssr: true });
```
mark-r-bjss, line 9: Does this need to be here or is this something that is global to the whole app and might be better in one of the `resource` files?

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+type NHSNotifyAutheticatorProps = {
+  children: ReactNode;
+};
+
+const isPathnameExempt = (pathname: string) => {
+  return (
+    pathname.includes('auth') ||
```
mark-r-bjss, line 17: This seems like it needs some abstraction and encapsulation, so that the page itself defines if it is exempt.

src/__tests__/components/forms/CreateSmsTemplate/CreateSmsTemplate.test.tsx
chris-elliott-nhsd: My PC was timing out doing the 300 characters and the focus of the unit test is that the character count component works - which can still be done with 30

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
chris-elliott-nhsd: the resource files aren't a part of the web app. this is code that every page has to run through as NHSNotifyAuthenticator is called from the layout

src/components/molecules/LoginStatus/LoginStatus.tsx
chris-elliott-nhsd: no, Amplify defines it as a type, not an enum
image

tests/accessibility/create-test-user.ts
mark-r-bjss: Can we also use an email that identifies us a bit better (in case somebody wonders where all these test emails come from), e.g. `nhs-notify-automated-accessibility-testing@nhs.net`

tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
```diff
@@ -33,19 +34,29 @@ export const nhsAppRadioSelectedSessionData: Session = {
   nhsAppTemplateMessage: '',
 };
 
+const testUserEmail = 'choose-page@nhs.net';
+const testUserPassword = 'Test-Password1';
```
mark-r-bjss, line 38: No hard coded passwords please (even for test).

tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
```diff
@@ -33,19 +34,29 @@ export const nhsAppRadioSelectedSessionData: Session = {
   nhsAppTemplateMessage: '',
 };
 
+const testUserEmail = 'choose-page@nhs.net';
```
mark-r-bjss, line 37: As per previous comment, lets include `nhs-notify-automated-test` as a prefix in case we end up sending something.

tests/test-team/template-mgmt-component-tests/template-mgmt-create-nhs-app-template-page.component.ts
```diff
@@ -13,17 +14,27 @@ export const nhsAppNoTemplateSessionData: Session = {
   nhsAppTemplateMessage: '',
 };
 
+const testUserEmail = 'create-nhs-app-page@nhs.net';
+const testUserPassword = 'Test-Password1';
```
mark-r-bjss, line 18: As above.

tests/test-team/template-mgmt-component-tests/template-mgmt-create-sms-template-page.component.ts
```diff
@@ -13,17 +14,27 @@ export const smsTemplateSessionData: Session = {
   nhsAppTemplateMessage: '',
 };
 
+const testUserEmail = 'create-sms-page@nhs.net';
+const testUserPassword = 'Test-Password1';
```
mark-r-bjss, line 18: As above.

tests/test-team/template-mgmt-component-tests/template-mgmt-start-page.component.ts
```diff
@@ -1,7 +1,21 @@
 import { test, expect } from '@playwright/test';
 import { TemplateMgmtStartPage } from '../pages/template-mgmt-start-page';
+import { TestUserClient } from '../helpers/test-user-client';
+
+const testUserEmail = 'start-page@nhs.net';
+const testUserPassword = 'Test-Password1';
```
mark-r-bjss, line 6: As above.

.gitleaksignore
mark-r-bjss: Didn't realise that they referred to other branches, that seems like a terrible feature, branches should be isolated from each other!

src/__tests__/components/forms/CreateSmsTemplate/CreateSmsTemplate.test.tsx
mark-r-bjss: Seems like there might be a more fundamental issue with the test if it times out.  Looks like we type then immediately query the page which may not have updated yet.  Reducing this to 30 does not seem like the right solution.  Maybe better to skip the test and add it into CCM-6950 to fix.

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
mark-r-bjss: Maybe the layout might be a better place for it.  It feels a bit more generic and not directly auth related.

mark-r-bjss: I'm thinking that we might need to add more config there in the future that is completely unrelated to this file.

src/components/molecules/Header/Header.tsx
chris-elliott-nhsd: i don't really understand what this is asking for!

tests/test-team/helpers/test-user-client.ts
chris-elliott-nhsd: we could do, but i'm not sure if it will change the behaviour very much - if we are trying to create a user that already exists, that is a problem, as the tests should be self-contained and responsible for their own data. so we should be throwing an error, which is what will happen now anyway if we try to create a user if it already exists

tests/accessibility/create-test-user.ts
chris-elliott-nhsd: updated - also while doing this realised that there could be a problem here with running the tests multiple times in parallel so i've added a UUID at the end of the email to avoid that

src/__tests__/components/forms/CreateSmsTemplate/CreateSmsTemplate.test.tsx
chris-elliott-nhsd: i don't think it's an issue with the test as it always passes on CI, it was just failing for me. not sure it's worth a specific ticket to look into something that only affects my PC. we can move it back to 300 if you feel this makes a significant difference to the test

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
chris-elliott-nhsd: i think that nextjs is asking us to make some tradeoffs here - if we want to ensure that by default a page is always protected by the authenticator, we need to put it in the layout code like this, where we do not have access to information about individual pages.

we can move the call to NHSNotifyAuthenticator outside of the layout code into specific page code, but then that comes with the obligation to remember to put it in every new page which i thought was less ideal than specifying exempt pages here. what do you think?

tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
chris-elliott-nhsd: updated

tests/test-team/template-mgmt-component-tests/template-mgmt-choose-page.component.ts
chris-elliott-nhsd: see other response

tests/test-team/template-mgmt-component-tests/template-mgmt-create-nhs-app-template-page.component.ts
chris-elliott-nhsd: see other response

tests/test-team/template-mgmt-component-tests/template-mgmt-create-sms-template-page.component.ts
chris-elliott-nhsd: see other response

tests/test-team/template-mgmt-component-tests/template-mgmt-start-page.component.ts
chris-elliott-nhsd: see other response

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
chris-elliott-nhsd: moved up to layout

src/__tests__/components/forms/CreateSmsTemplate/CreateSmsTemplate.test.tsx
chris-elliott-nhsd: changed back to 300

tests/test-team/helpers/test-user-client.ts
mark-r-bjss: What if the tests fail and leave the user behind?  I was thinking along the lines of self-cleaning tests, i.e. clean up after the last run before doing the new one.

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
mark-r-bjss: I'd go with default of "authenticated" and explicit "public" so that most of the time you don't need to do anything special.  Also happy to iterate on this, it's not a blocker.

src/components/molecules/Header/Header.tsx
mark-r-bjss: A named `const` for the data `auth`, e.g. `const LOGIN_PAGE_PATH = 'auth'`.  Helps the reader understand *why* it does what it does.

mark-r-bjss (dismissed):

src/__tests__/components/molecules/Header.test.tsx
```diff
@@ -1,12 +1,34 @@
...
+  it('auth path', async () => {
+    jest.mocked(usePathname).mockReturnValue('/auth');
+
+    const container = render(
+      <Authenticator.Provider>
+        <NHSNotifyHeader />
+      </Authenticator.Provider>
+    );
+
+    await waitFor(() => {});
```
bhansell1, line 18: why the empty waitfor?

src/__tests__/components/molecules/LoginStatus.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  fetchAuthSession: jest.fn(),
+}));
+
+test('LoginStatus - unauthenticated', async () => {
+  jest.mocked(useAuthenticator).mockReturnValue(
+    mockDeep<ReturnType<typeof useAuthenticator>>({
+      authStatus: 'unauthenticated',
+    })
+  );
+  jest.mocked(fetchAuthSession).mockResolvedValue({
```
bhansell1, line 22: Why not declare this mock once?

```
const fetchAuthSessionMock = jest.mocked(fetchAuthSession)
```

Also seems like good practice to call `after/beforeEach(jest.resetAllMocks)` to ensure mocks are cleared.

src/__tests__/components/molecules/NHSNotifyAuthenticator.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+  );
+  jest.mocked(usePathname).mockReturnValue('/redirect');
+  const redirectMock = jest.mocked(redirect);
+
+  render(
+    <NHSNotifyAuthenticator>
+      <p>testing</p>
+    </NHSNotifyAuthenticator>
+  );
+  await waitFor(() => {});
```
bhansell1, line 43: Same here, any need for the empty waitFor?

src/__tests__/components/molecules/Redirect.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+} from 'next/navigation';
+
+jest.mock('next/navigation', () => ({
+  ...jest.requireActual('next/navigation'),
+  redirect: jest.fn(),
+  useSearchParams: jest.fn(),
+}));
+
+test('Redirect - URL provided', () => {
+  const mockRedirect = jest.fn(mockDeep<typeof redirect>());
```
bhansell1, line 17: You shouldn't need to mock deep this function.

Doing;

```
const redirectMock = jest.mocked(redirect)
```

Should be fine. I've used the same pattern in other tests.

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable @typescript-eslint/no-require-imports,unicorn/prefer-module */
```
bhansell1, line 1: Are these needed? 

If they are and you think they're silly rules, then disable them globally?

src/__tests__/components/molecules/Header.test.tsx
chris-elliott-nhsd: it stops jest giving you warnings about wrapping everything in an `act()` block

src/__tests__/components/molecules/NHSNotifyAuthenticator.test.tsx
chris-elliott-nhsd: see other comment

src/__tests__/components/molecules/LoginStatus.test.tsx
chris-elliott-nhsd: updated

src/__tests__/components/molecules/Redirect.test.tsx
chris-elliott-nhsd: updated

src/components/molecules/NHSNotifyAuthenticator/NHSNotifyAuthenticator.tsx
chris-elliott-nhsd: these disables are not needed. in general, i think these are good rules, it's just unfortunate that there isn't a better way than require() to see the amplify outputs file

bhansell1 (approved):

sean-corrigan1 (approved):

PR #190, opened 2024-11-20
Draft: CCM-7409: add sandbox

harrim91:
## Type of changes

## Checklist

---

### DIFF
--- a/.github/workflows/stage-4-acceptance.yaml
+++ b/.github/workflows/stage-4-acceptance.yaml
@@ -60,13 +60,16 @@ jobs:
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/${{ secrets.AWS_ASSUME_ROLE_NAME }}
           role-session-name: deployInfra
           aws-region: ${{ env.AWS_REGION }}
-      - name: "Create sandbox"
+      - name: "Create Amplify sandbox"
         run: |
-          ./scripts/create_sandbox.sh
+          ./scripts/create_amplify_sandbox.sh
       - uses: actions/upload-artifact@v4
         with:
           name: amplify_outputs.json
           path: amplify_outputs.json
+      - name: "Create Backend sandbox"
+        run: |
+          ./scripts/create_backend_sandbox.sh wf-${GITHUB_RUN_ID}
   test-security:
     name: "Security test"
     runs-on: ubuntu-latest
--- a/infrastructure/terraform/components/sandbox/provider_aws.tf
+++ b/infrastructure/terraform/components/sandbox/provider_aws.tf
@@ -1,0 +1,24 @@
+provider "aws" {
+  region = var.region
+
+  allowed_account_ids = [
+    var.aws_account_id,
+  ]
+
+  default_tags {
+    tags = local.default_tags
+  }
+}
+
+provider "aws" {
+  alias  = "us-east-1"
+  region = "us-east-1"
+
+  default_tags {
+    tags = local.default_tags
+  }
+
+  allowed_account_ids = [
+    var.aws_account_id,
+  ]
+}
--- a/scripts/create_backend_sandbox.sh
+++ b/scripts/create_backend_sandbox.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# expect 1 argument to the script
+if [ $# -ne 1 ]; then
+  echo 1>&2 "$0: expected 1 arguments, received $#"
+  exit 2
+fi
+
+GROUP="nhs-notify-template-management-nonprod"
+AWS_REGION="eu-west-2" # make dynamic?
+identifier=$1
+
+echo "Creating backend sandbox \"$identifier\""
+
+
+cd $(git rev-parse --show-toplevel)/infrastructure/terraform
+
+./bin/terraform.sh \
+  --project nhs-notify \
+  --region $AWS_REGION \
+  --component sandbox \
+  --environment $identifier \
+  --group $GROUP \
+  --action apply \
+  -- \
+  -var-file="$(pwd)/etc/sandbox.tfvars" \
+  -var environment=$identifier
+
--- a/infrastructure/terraform/components/sandbox/.tool-versions
+++ b/infrastructure/terraform/components/sandbox/.tool-versions
@@ -1,0 +1,1 @@
+terraform 1.9.2
--- a/infrastructure/terraform/components/sandbox/locals_tfscaffold.tf
+++ b/infrastructure/terraform/components/sandbox/locals_tfscaffold.tf
@@ -1,0 +1,23 @@
+locals {
+  csi = replace(
+    format(
+      "%s-%s-%s",
+      var.project,
+      var.environment,
+      var.component,
+    ),
+    "_",
+    "",
+  )
+
+  default_tags = merge(
+    var.default_tags,
+    {
+      Project     = var.project
+      Environment = var.environment
+      Component   = var.component
+      Group       = var.group
+      Name        = local.csi
+    },
+  )
+}
--- a/infrastructure/terraform/components/sandbox/module_templates_api.tf
+++ b/infrastructure/terraform/components/sandbox/module_templates_api.tf
@@ -1,0 +1,11 @@
+module "templates_api" {
+  source = "../../modules/templates-api"
+
+  project               = var.project
+  environment           = var.environment
+  aws_account_id        = var.aws_account_id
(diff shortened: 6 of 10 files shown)


.github/workflows/stage-4-acceptance.yaml
```diff
@@ -172,6 +194,21 @@ jobs:
...
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/${{ secrets.AWS_ASSUME_ROLE_NAME }}
+          role-session-name: deployInfra
+          aws-region: eu-west-2
+      - name: "Destroy Backend sandbox"
         run: |
-          npm run destroy-sandbox -- --identifier "wf-${GITHUB_RUN_ID}"
+          npm run destroy-backend-sandbox wf-${GITHUB_RUN_ID}
```
chris-elliott-nhsd, line 214: you appear to have lost the teardown command for the Amplify sandbox

scripts/create_backend_sandbox.sh
```diff
@@ -0,0 +1,72 @@
...
+  echo 1>&2 "$0: expected 1 arguments, received $#"
+  exit 2
+fi
+
+ENVIRONMENT=$1
+COMPONENT="sandbox"
+AWS_ACCOUNT_ID="$(aws sts get-caller-identity --query Account --output text)"
+AWS_REGION="eu-west-2"
+PROJECT="nhs-notify"
+GROUP="nhs-notify-template-management-nonprod"
```
chris-elliott-nhsd, line 16: would this still work if we gave it a group that didn't clash with an existing group in the internal repo? seems safer to do that if possible

scripts/create_backend_sandbox.sh
```diff
@@ -0,0 +1,72 @@
...
+  }
+}";
+
+# write the backend file
+echo -e "${backend_config}" > backend_tfscaffold.tf
+
+# clean up the file on exit
+trap "rm -f $(pwd)/backend_tfscaffold.tf" EXIT;
+
+terraform init -upgrade
```
chris-elliott-nhsd, line 70: what does the extra init do?

.github/workflows/stage-4-acceptance.yaml
harrim91: good spot - fixed

scripts/create_backend_sandbox.sh
```diff
@@ -0,0 +1,72 @@
...
+
+terraform output -json > ${root_dir}/sandbox_tf_outputs.json
```
chris-elliott-nhsd, line 72: this looks good but not 100% clear how it works. how does this know where to find the terraform outputs? is that what the backend_tfscaffold stuff is doing?

scripts/create_backend_sandbox.sh
harrim91: Let me try it without 😅

harrim91: Yup, you are right. The problem was that the `terraform.sh` script creates and then deletes the backend config, so I had to manually create that again. But before I'd done that, the error I was seeing said about running `init` again. So I just added that in without really thinking it through. Thanks.

scripts/create_backend_sandbox.sh
harrim91: I'm not fully sure of everything `group` is for really - in this repo all I can see is it being used for tagging. In internal, it's used to pick the tfvars file. I think I added this when I was going down the route of trying to deploy this through the internal repo, and this would have been the right file to use. But I don't know if it's even needed here apart from for tagging.

scripts/create_backend_sandbox.sh
harrim91: Yes. The backend config block has the location of the tf state file, the `terraform output` command basically reads the state file and prints the outputs from there. There are other commands like `terraforms how` which you can use to interrogate specific resources. But this seemed like the right one to use here.

mark-r-bjss (dismissed):
Needs platform review as well.

scripts/create_backend_sandbox.sh
harrim91: Maybe worth seeing what platform folk think when they review it

chris-elliott-nhsd (dismissed):

.github/workflows/stage-4-acceptance.yaml
```diff
@@ -60,17 +60,39 @@ jobs:
...
-      - name: "Create sandbox"
+      - name: "Create Amplify sandbox"
         run: |
-          ./scripts/create_sandbox.sh
+          ./scripts/create_amplify_sandbox.sh
       - uses: actions/upload-artifact@v4
         with:
           name: amplify_outputs.json
           path: amplify_outputs.json
+  sandbox-set-up:
```
aidenvaines-bjss, line 70: I dont believe this will work. under the AMET deigns there is no ability for the component/microservice repositories to be able to do anything with the NHS Notify deployment context. The internal repo is where we manage how the services are deployed in the context of NHS Notify as the service.

@mark-r-bjss @timireland @RossBugginsNHS @m-houston and I discussed a mechanism by where there would/could be a workflow in the internal side of the programme whereby this sort of targeted creation and execution of service context testing could be run; though I don't know what happened there in my absence over the last few weeks

basically the service repos cant access AWS so this aint gunna work

aidenvaines-bjss (changes requested):

.github/workflows/stage-4-acceptance.yaml
aidenvaines-bjss: spoke to Mark R, this is a workaround

infrastructure/terraform/components/sandbox/provider_aws.tf
```diff
@@ -0,0 +1,24 @@
...
+  ]
+
+  default_tags {
+    tags = local.default_tags
+  }
+}
+
+provider "aws" {
+  alias  = "us-east-1"
+  region = "us-east-1"
```
mark-r-bjss, line 15: Is this correct?

infrastructure/terraform/modules/templates-api/variables.tf
```diff
@@ -55,8 +55,14 @@ variable "log_retention_in_days" {
...
   type = object({
-    USER_POOL_ID: string,
-    USER_POOL_CLIENT_ID: string
+    USER_POOL_ID : string,
+    USER_POOL_CLIENT_ID : string
   })
   description = "Cognito config"
 }
+
+variable "enable_sourcemaps" {
```
mark-r-bjss, line 64: I'm not sure we need a switch for this, not sure we would even switch it off.

infrastructure/terraform/modules/typescript-build-zip/null_resource_typescript_build.tf
```diff
@@ -5,6 +5,6 @@ resource "null_resource" "typescript_build" {
 
   provisioner "local-exec" {
     working_dir = var.source_code_dir
-    command     = "npm ci && npm run build"
+    command     = "rm -rf ${var.output_dir} && npm ci && npx esbuild --bundle --minify ${var.include_sourcemaps ? "--sourcemap=inline" : ""} --target=es2020 --platform=node --outdir=${var.output_dir} ${join(" ", var.entrypoints)}"
```
mark-r-bjss, line 8: Would it be nicer for the TF to call a bash script to allow this to be formatted nicely and to make it easier to extend in the future?

mark-r-bjss (approved):

infrastructure/terraform/modules/templates-api/variables.tf
harrim91: I was a bit confused about how to get it working properly - (I could only get it working with inline source maps, resulting in enormous files rather than a small file and a bigger source map) - so I thought it'd be good to have the option to turn it off. But I figured it out now, so have removed the flag.

infrastructure/terraform/modules/typescript-build-zip/null_resource_typescript_build.tf
harrim91: I think that would be more complicated than it's worth now - could do it in the future if this command did get more complicated. But I can't really see it needing to get more complex than this. And putting it in a bash script adds complexity about which directory it's running from, parsing arguments etc. Whereas it's working fine like this. 😅

mark-r-bjss (approved):

aidenvaines-bjss (approved):

bhansell1 (approved):

PR #197, opened 2024-11-26
CCM-7525: Add template API

bhansell1:
## Description

- Adds create, update, get and list API endpoints for templates.
- Adds new stage in test CI to check for differences in generated dependencies

## Type of changes

## Checklist

- [x] This PR is a result of pair or mob programming

---


tsconfig.json
```diff
@@ -32,7 +32,9 @@
       "@testhelpers": ["./src/__tests__/helpers"],
       "@utils/*": ["./src/utils/*"],
       "@domain/*": ["./src/domain/*"],
-      "@hooks/*": ["./src/hooks/*"]
+      "@hooks/*": ["./src/hooks/*"],
+      "@backend-api/templates/*": ["./lambdas/backend-api/src/templates/*"],
```
bhansell1, line 36: This is here to appease the linting lords. It shouldn't really be here. It should be getting resolved as part of https://nhsd-jira.digital.nhs.uk/browse/CCM-7005

lambdas/backend-api/src/templates/domain/template/template-repository.ts
```diff
@@ -0,0 +1,203 @@
...
+    TableName: process.env.TEMPLATES_TABLE_NAME,
+    Key: {
+      id: templateId,
+      owner,
+    },
+    UpdateExpression: `SET ${updateExpression.join(', ')}`,
+    ExpressionAttributeNames: expressionAttributeNames,
+    ExpressionAttributeValues: expressionAttributeValues,
+    ConditionExpression:
+      'attribute_exists(id) AND #templateStatus = :not_yet_submitted AND #templateType = :templateType',
```
bhansell1, line 149: Attribute exists is here because the `UpdateCommand` will also create a new record if it does not exist. Which is really counterintuitive!

lambdas/backend-api/src/templates/domain/template/template-repository.ts
```diff
@@ -0,0 +1,231 @@
...
+      new QueryCommand({
+        TableName: process.env.TEMPLATES_TABLE_NAME,
+        KeyConditionExpression: '#owner = :owner',
+        ExpressionAttributeNames: {
+          '#owner': 'owner',
+        },
+        ExpressionAttributeValues: {
+          ':owner': owner,
+        },
+        Limit: 50,
```
bhansell1, line 212: This is temporary. It's just to bring the same functionality to the API as the amplify db calls. We should probably look todo pagination or something similar in the future.

lambdas/backend-api/src/templates/domain/template/template-schema.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const $NhsAppTemplate = schemaFor<CreateTemplate>()(
+  $Template.extend({
+    templateType: z.literal(TemplateType.NHS_APP),
+    message: z
+      .string()
+      .min(1)
+      .max(MAX_NHS_APP_CHARACTER_LENGTH)
+      // eslint-disable-next-line security/detect-non-literal-regexp
+      .refine((s) => !new RegExp(NHS_APP_DISALLOWED_CHARACTERS, 'gi').test(s), {
```
bhansell1, line 40: if you use regex like `/[a-Z]/gi.test('pickle')`  as opposed to n`ew Regex('[a-Z]', 'gi).test('pickle')` .

The literal regex` /[a-Z]/` will sometimes pass/fail if you call it multiple times.

It's a bug in JS where the matching index is sometimes not reset and thus the wonky behaviour. Newing up a Regex on each call will always reset the matching indexes to 0 and will consistently work as expected.

lambdas/backend-api/src/utils/validate.ts
```diff
@@ -0,0 +1,43 @@
...
+      value.join(', '),
+    ])
+  );
+};
+
+export const validate = async <T extends z.Schema>(
+  $schema: T,
+  dto: unknown
+): Promise<ApplicationResult<z.infer<T>>> => {
+  const { error, data } = await $schema.safeParseAsync(dto);
```
bhansell1, line 31: So, I thought I had issues with some async issues in Zod due to the regex refine. So I used the `safeParseAsync` function to see if this helps. It didn't see https://github.com/NHSDigital/nhs-notify-web-template-management/pull/197/files#r1867773165.

Now I don't want to change it back (too many changes).

lambdas/backend-client/src/client.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const failure = (data: Failure) => ({
+  error: {
+    code: data.statusCode,
+    message: data.technicalMessage,
+    details: data.details,
+  },
+});
+
+export const templateClient: ITemplateClient = {
```
bhansell1, line 25: this whole file is just an experiment/example (hence no unit tests). The intention behind `ITemplateClient` is both the backend-api and the web-ui both implement the interface so that if any types, functions or signatures change then it'll break both sides.

infrastructure/terraform/modules/templates-api/module_get_template_lambda.tf
```diff
@@ -0,0 +1,52 @@
+module "get_template_lambda" {
+  depends_on  = [module.build_template_lambda, module.build_template_client]
```
chris-elliott-nhsd, line 2: you shouldn't need depends_on for resources that are explicitly referenced in the resource

infrastructure/terraform/modules/templates-api/spec.tmpl.json
```diff
@@ -58,9 +273,169 @@
...
+          "message": {
+            "type": "string"
+          },
+          "subject": {
+            "type": "string"
+          },
+          "templateType": {
+            "$ref": "#/components/schemas/TemplateType",
+            "readOnly": true,
+            "description": "This value will never be updated. It is used to determine the type of template being validated."
```
chris-elliott-nhsd, line 386: if this is never updated we probably don't want it in the API request - the API can figure out for itself how to validate an update to an already existing template without the user telling them

infrastructure/terraform/modules/templates-api/variables.tf
```diff
@@ -61,3 +61,20 @@ variable "cognito_config" {
   description = "Cognito config"
 }
 
+variable "template_api_entrypoints" {
```
harrim91, line 64: What is the use-case for this being a variable?

infrastructure/terraform/modules/templates-api/spec.tmpl.json
bhansell1: The purpose of it being here is use the correct zod schema when validating the template. 

If we remove this field then I'd need to fetch the item from the database before I do the update or split the API into /templates/email,sms,app. 

My thinking was this was a seemingly decent middle ground. What do you think?

infrastructure/terraform/modules/templates-api/module_get_template_lambda.tf
bhansell1: Oh i didn't know that! I'll try it out!

infrastructure/terraform/modules/templates-api/variables.tf
bhansell1: I couldn't think of a better way to put all the entrypoints I need in one place that is also be used in of the modules. 

Please do suggest a better way.

infrastructure/terraform/modules/templates-api/module_build_template_lambda.tf
```diff
@@ -0,0 +1,18 @@
...
+  source_code_dir = "${local.lambdas_source_code_dir}/backend-api"
+  entrypoints     = [
+    var.template_api_entrypoints.create_template,
+    var.template_api_entrypoints.get_template,
+    var.template_api_entrypoints.update_template,
+    var.template_api_entrypoints.list_template,
+  ]
+}
+
+module "build_template_client" {
```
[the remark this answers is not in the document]
bhansell1: I need this to build `backend-client`. Which `backend-api` relies on. The reason for this is explained in the `backend-client` [README.md.](https://github.com/NHSDigital/nhs-notify-web-template-management/pull/197/files#diff-ad92c6f5c10ae6ba941008f698aa7db777b5d60f691b42d2b26aebe910130b37)

lambdas/backend-api/src/utils/schema-for.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from 'zod';
+
+export const schemaFor =
+  <Output, Input = Output>() =>
+  <S extends z.ZodType<Output, z.ZodTypeDef, Input>>(schema: S) =>
+    schema;
```
chris-elliott-nhsd, line 6: isn't this just longhand for z.infer?

lambdas/backend-api/src/templates/app/get-template.ts
```diff
@@ -0,0 +1,38 @@
+import { ITemplateClient } from 'nhs-notify-backend-client';
+import { templateRepository } from '@backend-api/templates/domain/template';
+import { userRepository } from '@backend-api/templates/domain/user';
+import { logger, success } from '@backend-api/utils/index';
+
+export const getTemplate: ITemplateClient['getTemplate'] = async (
+  templateId,
+  token
+) => {
+  const userResult = await userRepository.getUser(token);
```
harrim91, line 10: Instead of doing this, you can get the authorizer to do this (the authorizer does do this already in fact), and add it to the request context (it doesn't do this bit). Let me dig something out on the subject.

bhansell1: That would be great! I thought we could but didn't look into it TBH. 

IIRC, you can return context variables from the authorizer lambda.

lambdas/backend-api/src/utils/schema-for.ts
bhansell1: I didn't want to use z.infer because the types are autogenerated. So I needed to way to enforce type correctness with z schemas. This is entirely ripped from comms-mgr. 

I think it's pretty neat TBH!

infrastructure/terraform/modules/templates-api/module_get_template_lambda.tf
bhansell1: Ah, so I just tried removing the depends_on but I got a bunch of errors. Might be one to talk through tomorrow!

lambdas/backend-api/src/templates/app/get-template.ts
harrim91: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html

Basically the authorizer lambda already checks the JWT and even calls Cognito to check that the user exists.

In the return value of the authorizer, you can return something like this:

```ts
return {
  principalId: 'api-caller',
  policyDocument: {
    Version: '2012-10-17',
    Statement: [
      {
        Action: 'execute-api:Invoke',
        Effect,
        Resource,
      },
    ],
  },
  context: {
    user: JSON.stringify({ id: user.id })
  }
}
```

That `context` object gets passed downstream by API Gateway into the handler lambda. And you can use it like `event.requestContext.authorizer.user`

It's not a massive thing I don't think. Because you aren't actually making a network call to Cognito or anything to get the user, just parsing it from the JWT. But having it in the authorizer means you can have the same logic once, and all your auth/jwt logic in one place where it belongs. Rather than doing it in each Lambda, and the actual endpoint lambdas don't have to concern themselves with JWTs and other auth-related things.

harrim91: I guess you'd still want to Zod your `event.requestContext.authorizer.user` - but you're doing that anyway

.eslintrc.json
```diff
@@ -73,7 +73,8 @@
     "unicorn/no-null": "off",
     "prefer-regex-literals": "off",
     "class-methods-use-this": "off",
-    "no-underscore-dangle": "off"
+    "no-underscore-dangle": "off",
+    "import/no-extraneous-dependencies": ["error", { "devDependencies": true }]
```
harrim91, line 77: What is this changed for?

lambdas/backend-api/src/templates/domain/template/template-repository.ts
```diff
@@ -0,0 +1,231 @@
...
+      new PutCommand({
+        TableName: process.env.TEMPLATES_TABLE_NAME,
+        Item: entity,
+        ReturnConsumedCapacity: 'TOTAL',
+      })
+    );
+
+    const consumedUnits = response?.ConsumedCapacity?.CapacityUnits || 0;
+
+    // Note: This is a bit strange, PutCommand does not return the any data when creating
```
harrim91, line 80: Can you not just assume it worked if it didn't throw an error?

lambdas/backend-api/src/templates/domain/user/user.ts
```diff
@@ -0,0 +1,3 @@
+export type User = {
+  id: string;
```
harrim91, line 2: Why is this `id` not `username` (or the other way around)? `type User` doesn't match `$User` - without the mapping you could do `schemaFor<User>` in the user-schema above too.

lambdas/backend-client/src/client.ts
harrim91, line 25: Could this be a class? Put the axios instance as a class member, and pass the token into the constructor? 

I can see how this breaks the backend/frontend thing. And why you've got the token going into the backend methods and are doing the token parsing there. 

It'd be nice if you could instantiate the api client used in the ui with a token, and just pass the template data into `createTemplate` etc. And the backend implementation of `createTemplate` would have the same signature, and the implementation would be instantiated with `event.requestContext.authorizer`. 

This might be getting to be a big lot of changes I'm asking for though 😅

infrastructure/terraform/modules/templates-api/spec.tmpl.json
mark-r-bjss: We'll be fetching the data anyway to do an authorisation check, i.e. are you allowed to update it?

lambdas/backend-client/src/client.ts
```diff
@@ -0,0 +1,101 @@
...
+      },
+    });
+
+    if (response.data.statusCode !== 200) {
+      return failure(response.data as Failure);
+    }
+
+    return {
+      statusCode: response.data.statusCode,
+      data: (response.data as SuccessList).items,
```
chris-elliott-nhsd: the typecast here is unnecessary, you can have client.get<SuccessList>m also in many other places

lambdas/backend-api/src/templates/domain/template/template-repository.ts
```diff
@@ -0,0 +1,231 @@
...
+    '#updatedAt = :updateAt',
+    '#templateStatus = :templateStatus',
+  ];
+
+  let expressionAttributeNames: Record<string, string> = {
+    '#name': 'name',
+    '#message': 'message',
+    '#templateStatus': 'templateStatus',
+    '#updatedAt': 'updatedAt',
+    '#templateType': 'templateType',
```
chris-elliott-nhsd, line 116: most of these are not protected words so the use of expressionAttributeNames is not needed

lambdas/backend-api/src/__tests__/templates/domain/template/template-repository.test.ts
```diff
@@ -0,0 +1,399 @@
...
+  QueryCommand,
+  UpdateCommand,
+} from '@aws-sdk/lib-dynamodb';
+import { mockClient } from 'aws-sdk-client-mock';
+import {
+  TemplateStatus,
+  TemplateType,
+  UpdateTemplate,
+} from 'nhs-notify-backend-client';
+import { v4 as uuidv4 } from 'uuid';
```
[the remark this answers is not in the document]
bhansell1: It should be node:crypto. 

I've updated it.

lambdas/backend-api/src/templates/app/get-template.ts
bhansell1: Updated. to just pass the username in the authorizer context.

lambdas/backend-api/src/templates/domain/user/user.ts
bhansell1: I've removed User completely.

lambdas/backend-client/src/client.ts
bhansell1: Done. 

Good suggestion, thanks!

.eslintrc.json
bhansell1: It was due to the @types/uuid stating that it _shouldnt_ be a devDepency. This line and uuid have both been removed.

lambdas/backend-api/src/templates/domain/template/template-repository.ts
bhansell1: Probably? 

I just think it's weird that dynamo can _not_ writing something to the DB and still return a success while not actually saving the record.

lambdas/backend-client/src/client.ts
bhansell1: I need this typecasting to access the .items value and similarly the .template property from the API response.

chris-elliott-nhsd: the formatting on the previous comment didn't come out right but when you are specifying the type on the client.get line you just need to specify the right type and it will be unnecessary

chris-elliott-nhsd (approved):

lambdas/backend-api/src/templates/domain/template/template-repository.ts
bhansell1: removed

infrastructure/terraform/modules/templates-api/spec.tmpl.json
bhansell1: Chatted on teams about this.

Decided leaving this as-is is a good enough middle ground.

infrastructure/terraform/modules/templates-api/variables.tf
harrim91: Just don't know the use-case why they would need to be exposed as parameters. Could use locals

```tf
locals {
  template_api_entrypoints = {
    create_template = "src/templates/api/create.ts"
    get_template = "src/templates/api/get.ts"
  }
}
```

Then reference them like `local.template_api_entrypoints.create_template`

infrastructure/terraform/modules/templates-api/module_get_template_lambda.tf
bhansell1: Terraform tells me I need to use `depends_on` others a `circular dependency` error is thrown by Terraform.  

I want to build the `backend-api` and `backend-client` once rather than building multiple times per lambda.

infrastructure/terraform/modules/templates-api/variables.tf
bhansell1: Updated. Thanks for the suggestion.

lambdas/backend-client/src/client.ts
bhansell1: I've updated this. 

We're no longer casting and using the specific type for each api call.

infrastructure/terraform/modules/templates-api/module_build_template_lambda.tf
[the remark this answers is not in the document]
bhansell1: If I don't run the typescript build on the `backend-client` then I get build failures because `backend-client` has a dependency on `axios`. 

If I don't run the `typescript-build` on this package then the decencies for `backend-client` are not installed and we get a build failure stating that `axios` cannot be found.  

This is because `backend-client` is a local package rather than an external package. 

image

harrim91 (dismissed):

chris-elliott-nhsd (approved):

mark-r-bjss (approved):

chris-elliott-nhsd (approved):

Kennedy1265 (approved):

PR #208, opened 2024-12-06
CCM-7462: add login logout pages on dev server

harrim91:
## Description

Adds sign in and sign out pages to templates app which are only available when running the dev server.
Makes whole app auth-aware by wrapping in Amplify Auth Provider.

## Context

Allows us to authenticate when running locally without the auth web app available.

## Type of changes

## Checklist

---


next.config.js
```diff
@@ -24,6 +27,12 @@ const nextConfig = {
       },
     ];
   },
-};
+
+  // pages with e.g. .dev.tsx extension are only included when running locally
+  pageExtensions: ['ts', 'tsx', 'js', 'jsx'].flatMap((extension) => {
```
chris-elliott-nhsd, line 32: this does not work because .dev.tsx extensions are still .tsx extensions, so the dev pages will still get deployed everywhere

harrim91: It works for me. I deleted `frontend/.next` directory, ran `npm run build` to create a production build and then `npm start` to run the production build. The login and logout pages weren't available and I got a 404 instead.

chris-elliott-nhsd: wow. my mistake! i feel like i tried to do this before and ended up having to rename all the other pages but if this works fine

frontend/amplify/package.json
```diff
@@ -4,6 +4,14 @@
   "private": true,
   "main": "index.ts",
   "type": "module",
+  "scripts": {
+    "generate-dependencies": "tsx functions/send-email/generate-dependencies.ts",
+    "lint": "eslint . -c .eslintrc.json",
+    "lint:fix": "eslint . --fix",
+    "test:unit": "jest",
+    "typecheck": "tsc --noEmit",
+    "typescript": "^5.5.4"
```
bhansell1, line 13: How comes this is in scripts and not dependencies?

frontend/src/app/auth/page.dev.tsx
```diff
@@ -0,0 +1,40 @@
+'use client';
+
+import React, { Suspense } from 'react';
+import { useSearchParams, redirect, RedirectType } from 'next/navigation';
+import { Authenticator } from '@aws-amplify/ui-react';
+import '@aws-amplify/ui-react/styles.css';
+// this is an npm module, not node:path
+// eslint-disable-next-line  unicorn/prefer-node-protocol
+import path from 'path';
```
bhansell1, line 9: path is in node. Why the change? 

https://nodejs.org/docs/latest-v18.x/api/path.html#pathnormalizepath

lambdas/endpoint/dist/index.js
bhansell1, line 1: I don't think endpoint should still be here?

frontend/amplify/package.json
harrim91: 🤔

harrim91: Good spot. It was already in `scripts` and I just moved the entire `scripts` block. But I will fix it.

frontend/src/app/auth/page.dev.tsx
harrim91: This `Redirect` component was copied from the auth webapp - https://github.com/NHSDigital/nhs-notify-iam-webauth/blob/main/src/components/molecules/Redirect/Redirect.tsx

I'm guessing an npm package is being used because it's a client component so doesn't have access to Node core modules.

lambdas/endpoint/dist/index.js
harrim91: 😳

frontend/amplify/package.json
harrim91: fixed

frontend/src/app/auth/page.dev.tsx
bhansell1: ahhhhh fair!

chris-elliott-nhsd (approved):

lambdas/backend-api/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.1",
   "scripts": {
     "generate-dependencies": "tsx scripts/generate-dependencies.ts",
-    "test:unit": "jest",
+    "test:unit": "npm run generate-dependencies && jest",
```
chris-elliott-nhsd, line 7: if we're running generate dependencies as part of the test:unit commands, we don't need to run it separately on the pipeline right? no strong views which one we keep but feels like we could save time by not having both

bhansell1 (approved):

chris-elliott-nhsd (approved):

lambdas/backend-api/package.json
harrim91: I added it here because I ran the test script locally and it failed because I hadn't run the script. I think it was the first time I'd run it since the workspaces thing came in and the file structure was changed. I added it because you need the dependency to make sure the tests pass, so it made sense to make the test command run it. 

But I've just realised that the linting and typecheck also fail if you don't have the `email-template.json` file. So with this approach, really I should be prefixing all of those commands with `generate-dependencies`. Which does seem a bit pointless - it'd be better to stick with the current approach and run it once manually when needed. 

And also there are a few packages that have the `generate-dependencies` script, and the pipelines are running it in all the workspaces. Whereas this is just in the `backend-api` package. So we'd still need to keep it in the pipeline for those other packages possibly.

So in conclusion, I will remove this!

harrim91: Updated

chris-elliott-nhsd (approved):

PR #264, opened 2025-01-15
CCM-5340: Add CSP header with nonce for next generated inline scripts

ClareJonesBJSS:
## Description

Adds CSP to headers via middleware

## Context

Reflects CSP added to web-gateway at https://github.com/NHSDigital/nhs-notify-web-gateway/pull/41, but nextjs generates inline scripts, so the header needs to be generated at this level so a nonce can be added

## Type of changes

## Checklist

---


frontend/next.config.js
```diff
@@ -20,7 +20,11 @@ const nextConfig = (phase) => {
 
     experimental: {
       serverActions: {
-        allowedOrigins: [domain, domain.replace('templates', 'web-gateway')],
+        allowedOrigins: [
+          domain,
+          domain.replace('templates', 'web-gateway'),
+          'alnu1.web-gateway.dev.nhsnotify.national.nhs.uk',
```
alexnuttall, line 26: remove before merge

Ian-Hodges (dismissed):

frontend/src/app/layout.tsx
```diff
@@ -9,6 +9,8 @@ export const metadata: Metadata = {
   description: content.global.mainLayout.description,
 };
 
+export const dynamic = 'force-dynamic';
```
bhansell1, line 12: How comes this is needed in the layout?

frontend/src/middleware.ts
```diff
@@ -2,15 +2,56 @@ import { NextResponse, type NextRequest } from 'next/server';
...
+    'default-src': [`'none'`],
+    'frame-ancestors': [`'none'`],
+    'font-src': [`'self'`, 'https://assets.nhs.uk'],
+    'form-action': [`'self'`],
+    'frame-src': [`'self'`],
+    'connect-src': [`'self'`, 'https://cognito-idp.eu-west-2.amazonaws.com'],
+    'img-src': [`'self'`],
+    'manifest-src': [`'self'`],
+    'object-src': [`'none'`],
+    'script-src': [`'self'`, `'nonce-${nonce}'`],
```
bhansell1, line 17: There are 2 scripts being loaded in `Layout.tsx`, are they being executed with this policy? If not you might need to add the `nonce` attribute to them where the value matches this generated nonce value.

Unless next is doing some weird behind the scenes shenanigans?

frontend/src/middleware.ts
```diff
@@ -2,15 +2,56 @@ import { NextResponse, type NextRequest } from 'next/server';
...
+    'style-src': [`'self'`, `'nonce-${nonce}'`],
+    'upgrade-insecure-requests;': [],
+  };
+
+  if (process.env.NODE_ENV === 'development') {
+    contentSecurityPolicyDirective['script-src'].push(`'unsafe-eval'`);
```
bhansell1, line 23: Out of interest what caused this to be added just for development?

frontend/src/middleware.ts
```diff
@@ -2,15 +2,56 @@ import { NextResponse, type NextRequest } from 'next/server';
...
 }
 
 export async function middleware(request: NextRequest) {
-  const excludedPaths = ['/create-and-submit-templates', '/auth'];
+  const nonce = Buffer.from(crypto.randomUUID()).toString('base64');
+
+  const csp = getContentSecurityPolicy(nonce);
+
+  const requestHeaders = new Headers(request.headers);
+  requestHeaders.set('Content-Security-Policy', csp);
```
bhansell1, line 41: I wonder if we can set this CSP via the web-gateway responses? Then we might not need to have this defined in both the iam and templates repo?

bhansell1: Oh wait they are set in the web gateway. Why do we also need to define them here?

frontend/src/middleware.ts
alexnuttall: Disabling it breaks hot reload and source maps (done via webpack)

frontend/src/middleware.ts
alexnuttall: The nonce needs to be generated for each request, so we're not able to set `script-src` using a static header policy

bhansell1: Could we use an edge lambda to do exactly that and generate a new nonce on each request and return it on the response headers?

alexnuttall: The second last comment [here](https://github.com/aws-samples/amazon-cloudfront-functions/issues/3) makes it sound like the body is not available in the response hook - would be good to experiment though

Ian-Hodges (approved):

mark-r-bjss (approved):

PR #332, opened 2025-03-06
CCM-8579: Switch from util types to api client types

ClareJonesBJSS:
## Description

Rewrites the API spec to take advantage of anyOf/oneOf to avoid having all channel-specific properties as optional properties on a single object, and removes util types in favour of the generated ones.

- Switches generators as the current one is unsupported and had some bugs I could only work around by structuring things weirdly and/or a post generation transform step
- Switches to unions over enums

## Context

On creating the letter create/update pages, I went through a lot of silly mistakes and stray properties that TS could have caught, along with little discrepancies where the generated types and nhs-notify-web-template-management-utils types/validations had diverged. Consolidating on TemplateDto gives a lot more confidence that types actually reflect reality at runtime and will be accepted by the API

## Testing

Manually tested SMS/email/NHS app template creation, editing, copying, deleting and submitting and all executed successfully with no visible regression. Letters seem to be working well so far for my follow on work and Alex's backend work at https://github.com/NHSDigital/nhs-notify-web-template-management/pull/340

Screenshot 2025-03-07 at 12-30-28 NHS Notify - Template Management
Screenshot 2025-03-07 at 12-31-36 NHS Notify - Template Management
Screenshot 2025-03-07 at 12-31-58 NHS Notify - Template Management
Screenshot 2025-03-07 at 12-32-25 NHS Notify - Template Management
Screenshot 2025-03-07 at 12-32-39 NHS Notify - Template Management
Screenshot 2025-03-07 at 12-33-08 NHS Notify - Template Management
Screenshot 2025-03-12 at 17 01 26

## Type of changes

## Checklist

---


frontend/src/components/forms/NhsAppTemplateForm/server-action.ts
```diff
@@ -43,7 +43,7 @@ export async function processFormActions(
   };
 
   const savedTemplate = await ('id' in updatedTemplate
-    ? saveTemplate(updatedTemplate)
+    ? saveTemplate(updatedTemplate.id, updatedTemplate)
```
Ian-Hodges, line 46: Does it need the id passed in separately? I can't see it being used where the id isn't from the template being passed in as the second parameter.

ClareJonesBJSS: I passed it separately as UpdateTemplate type has no id and I was trying to reflect more what the api was consuming as I was picking through, happy to change it back

lambdas/backend-client/src/schemas/template-schema.ts
```diff
@@ -1,37 +1,99 @@
...
+  );
+export type ValidatedTemplateDto = TemplateDto &
+  (
+    | EmailPropertiesWithType
+    | NhsAppPropertiesWithType
+    | SmsPropertiesWithType
+    | LetterPropertiesWithType
+  );
+
+const [firstTemplateType, ...remainingTemplateTypes] = TEMPLATE_TYPE_LIST;
```
alexnuttall, line 62: I think we could avoid this by using a util from core for the union lists

```
/**
 * Returns an identity function which will fail to compile if 'array'
 * doesn't contain all the cases of 'Union'
 *
 * @example
 * const arrayOfFooBarBaz = arrayOfAll<'foo' | 'bar' | 'baz'>();
 *
 * const a = arrayOfFooBarBaz(['foo', 'bar']); // does not compile
 * const b = arrayOfFooBarBaz(['foo', 'bar', 'baz']); // compiles
 */
export function arrayOfAll<Union>() {
  return <T extends [Union, ...Union[]]>(
    array: T & ([Union] extends [T[number]] ? unknown : 'Invalid')
  ) => array;
}
```
It's slightly modified from the version in core so T is a non-zero tuple. It doesn't prevent duplicates unfortunately, unlike the Record approach - but that seems OK for the `z.enum` case

frontend/package.json
```diff
@@ -20,14 +20,15 @@
...
+    "@aws-sdk/client-ses": "^3.637.0",
     "aws-amplify": "^6.12.3",
+    "clsx": "^2.1.1",
     "date-fns": "^4.1.0",
     "jsonwebtoken": "^9.0.2",
     "jwt-decode": "^4.0.0",
     "markdown-it": "^13.0.1",
     "mimetext": "^3.0.24",
-    "next": "^15.2.1",
+    "next": "^15.1.7",
```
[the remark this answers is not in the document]
ClareJonesBJSS: Agh, no that keeps happening and I'm not sure why, thanks for the spot

alexnuttall (approved):

chris-elliott-nhsd (approved):

gparry333 (approved):

PR #340, opened 2025-03-10
CCM-8585 upload pdf API

alexnuttall:
## Description
Adding a `POST -> /letter-template` endpoint to create letter templates.

The template is first updated to `PENDING_UPLOAD`, then one or both files are sent to s3, then the template is updated to `PENDING_VALIDATION`. 

Further downstream, we will have to wait and retry if the status hasn't reached  `PENDING_VALIDATION` yet.

Update endpoint has not been added, ticketed separately: https://nhsd-jira.digital.nhs.uk/browse/CCM-8993

## Type of changes

## Checklist

---


infrastructure/terraform/modules/backend-api/module_s3bucket_quarantine.tf
```diff
@@ -0,0 +1,55 @@
...
+  policy_documents = [
+    data.aws_iam_policy_document.s3bucket_pdf_template_quarantine.json
+  ]
+
+  default_tags = {
+    Name = "Quarantine for files pending virus scan"
+  }
+}
+
+data "aws_iam_policy_document" "s3bucket_pdf_template_quarantine" {
```
harrim91, line 35: Not sure this is needed - it looks like the same policy is included in the module anyway - https://github.com/NHSDigital/nhs-notify-shared-modules/blob/main/infrastructure/modules/s3bucket/data_iam_policy_document.tf

infrastructure/terraform/modules/backend-api/kms_key_s3.tf
```diff
@@ -0,0 +1,5 @@
+resource "aws_kms_key" "s3" {
```
harrim91, line 1: I've been wondering if we get any benefit from having so many kms keys. Would it not be sufficient to just use the key in `infrastructure/terraform/components/app/module_kms.tf`? Should probably see what Aiden thinks

harrim91: I know this is just the way we've always done it from the core project, maybe not the place to pull you up on this. 

My thinking is that KMS costs money and the data that we keep in SQS, S3, DynamoDB etc is all the same data. I'm not sure what benefit we get from having a different key per service, we could just have one key for the whole application right?

If we needed other keys, it feels like it should be on grounds of like data security - e.g. use a more restricted key for encrypting PII data.

infrastructure/terraform/modules/backend-api/module_s3bucket_quarantine.tf
alexnuttall: removed, thanks

infrastructure/terraform/modules/backend-api/kms_key_s3.tf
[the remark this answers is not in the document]
alexnuttall: removed

utils/test-helper-utils/src/pdf-letter-multipart.ts
```diff
@@ -0,0 +1,50 @@
...
+        typeof template === 'string' ? template : JSON.stringify(template)
+      );
+    }
+
+    if (part._type === 'file') {
+      fd.append(part.partName, part.file, {
+        filename: part.fileName,
+        contentType: part.fileType,
+      });
+    }
```
chris-elliott-nhsd, line 39: is it worth flagging unrecognised `_type` values?

lambdas/backend-client/src/types/template-client.ts
```diff
@@ -2,14 +2,25 @@ import { TemplateDto, CreateTemplate, UpdateTemplate } from './generated';
 import { Result } from './result';
 
 export interface ITemplateClient {
-  createTemplate(template: CreateTemplate): Promise<Result<TemplateDto>>;
```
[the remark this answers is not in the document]
ClareJonesBJSS: Yeah I didn't want to add sneaky extra refactors onto the PR preceding this, but lambdas/backend-api/src/templates/app/template-client.ts vs lambdas/backend-client/src/template-api-client.ts trips me up every time, I wouldn't even call the former a client really

infrastructure/terraform/modules/backend-api/spec.tmpl.json
```diff
@@ -455,33 +497,158 @@
...
-            "type": "string"
-          },
           "message": {
             "type": "string"
+          }
+        }
+      },
+      "LetterProperties": {
+        "type": "object",
+        "required": [
```
ClareJonesBJSS, line 513: Is it worth splitting this into BaseLetterProperties for and DtoLetterProperties so that files can be required in the dto but absent in the create?

alexnuttall: Done, although to be honest, it seems like there is de-duplication to be done with the validators

infrastructure/terraform/components/sandbox/module_kms.tf
```diff
@@ -0,0 +1,46 @@
...
+  region         = var.region
+
+  name                 = "main"
+  deletion_window      = var.kms_deletion_window
+  alias                = "alias/${local.csi}"
+  key_policy_documents = [data.aws_iam_policy_document.kms.json]
+  iam_delegation       = true
+}
+
+data "aws_iam_policy_document" "kms" {
```
harrim91, line 17: Don't think this is needed on the sandbox at the moment - the one in the API component is used for more stuff

infrastructure/terraform/modules/backend-api/variables.tf
```diff
@@ -47,6 +47,11 @@ variable "csi" {
   description = "CSI from the parent component"
 }
 
+variable "csi_global" {
```
harrim91, line 50: I can't see where this is being used on any resources

lambdas/backend-api/src/templates/api/create-letter.ts
```diff
@@ -0,0 +1,53 @@
...
+    const user = event.requestContext.authorizer?.user;
+
+    if (!user) {
+      return apiFailure(400, 'Invalid request');
+    }
+
+    const base64body = Buffer.from(event.body ?? '', 'base64');
+
+    const contentType =
+      event.headers['Content-Type'] ?? event.headers['content-type'] ?? 'none';
```
harrim91, line 24: Just out of curiosity. Have you seen the header come through with different casing? I know it's case-sensitive, but particularly for standard headers I thought they came through `Like-This` from APIG. Not a criticism or anything, I'm just curious.

alexnuttall: Just when sending via curl, see `lambdas/backend-api/README.md`

harrim91: Okay interesting that if you set it explicitly in the client, APIG doesn't standardize the casing. It's definitely good to have some handling for different cases.

ClareJonesBJSS (approved):

ClareJonesBJSS (approved):

infrastructure/terraform/modules/backend-api/spec.tmpl.json
```diff
@@ -210,6 +210,85 @@
...
+                  "properties": {
+                    "template": {
+                      "$ref": "#/components/schemas/CreateTemplate"
+                    },
+                    "letterPdf": {
+                      "type": "string",
+                      "format": "binary"
+                    },
+                    "testCsv": {
+                      "type": "string"
```
mark-r-bjss, line 239: Doesn't this need the `"format": "binary"` as well?

alexnuttall: I don't think the CSV part actually gets base64 encoded, let me check

alexnuttall: yep, it looks like this

Screenshot 2025-03-17 at 09 29 15

mark-r-bjss (dismissed):

gparry333 (dismissed):

mark-r-bjss (approved):

bhansell1 (approved):

PR #391, opened 2025-04-08
CCM-8574: letter file validation

harrim91:
## Description

Adds a Lambda which parses and validates uploaded letter template pdfs and accompanying test data csv files.

The lambda uses `pdfjs-dist`, which required adding a Lambda Layer with the right binaries to use the library. 

Simplifies some of the architecture of the previous virus scan work - removes the EB pipe and just parses the file data from the s3 object key. Everything just works directly from the GuardDuty scan result event. 

The validation lambda is fed from an SQS queue with a delivery delay configured. This gives the file copy and scan result database update a better chance of completing before the validation is attempted.

## Type of changes

## Checklist

- [x] I have updated the documentation accordingly

---


tests/test-team/fixtures/pdf-upload/with-personalisation/test-data.csv
```diff
@@ -1,5 +1,4 @@
-﻿parameter,example 1,example 2,example 3
-appointment_date,Wednesday 10 September 2025,Saturday 10 April 2025,Monday 1 May 2025
-appointment_time,12:56pm,11:56am,1:56pm
-appointment_location,"City, Sandwell & Walsall BSS, The Rosewood Centre, Sandwell & West Birmingham Hospitals NHS Trust, The Birmingham Treatment Centre, City Hospital, Dudley Road, Birmingham, B18 7QH","The Royal Shrewsbury Hospital, Breast Screening Office, Treatment Centre, Mytton Oak Road, Shrewsbury, SY3 8XQ","The Epping Breast Screening Unit, St Margaret's Hospital, The Plain, Epping, Essex, CM16 6TN"
-contact_telephone_number,020 3299 9010,020 3299 9010,020 3299 9010
+﻿appointment_date,appointment_time,appointment_location,contact_telephone_number
```
alexnuttall, line 1: I think this needs to stay in sync with the [published guidance](https://notify.nhs.uk/using-nhs-notify/personalisation) (under 'Providing example data'). @mark-r-bjss agrees this should stay as it is

infrastructure/terraform/modules/backend-api/cloudwatch_event_rule_guardduty_quarantine_scan_no_threats.tf
```diff
@@ -0,0 +1,47 @@
+resource "aws_cloudwatch_event_rule" "guardduty_quarantine_scan_no_threats" {
```
chris-elliott-nhsd, line 1: aren't `aws_cloudwatch_event_rule.guardduty_quarantine_scan_no_threats` and `aws_cloudwatch_event_rule.virus_scan_passed` essentially the same event? they're responding either directly or indirectly to a successful virus scan? why not just have them both be one event?

more generally, is there a particular reason not just to have a single event_rule/lambda that responds to the guard duty result instead of 4 separate event_rule/lambda combinations? is it just that it makes the individual lambdas simpler?

infrastructure/terraform/modules/lambda-function/variables.tf
```diff
@@ -53,8 +53,20 @@ variable "memory_size" {
   default     = 128
 }
 
+variable "timeout" {
+  description = "The amount of time your Lambda Function has to run in seconds."
+  type        = number
+  default     = 3
```
chris-elliott-nhsd, line 59: this seems very low! like any dynamo latency at all could potentially cause the lambdas to timeout?

infrastructure/terraform/modules/lambda_layer/null_resource_package_layer.tf
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  provisioner "local-exec" {
+    working_dir = var.source_code_dir
+    command     = "npm install --force"
+  }
+
+  provisioner "local-exec" {
+    working_dir = var.source_code_dir
+    command     = "cp -r node_modules/* $OUTPUT_DIR/layer/nodejs/node_modules"
```
chris-elliott-nhsd, line 29: we're trying to move towards a model where terraform isn't building the lambdas (https://nhsd-confluence.digital.nhs.uk/spaces/RIS/pages/1031451379/REFCOM-2025-021+%F0%9F%9F%A2+AMET+Workflow+for+Lambda+packaging)

could this be done in `pre.sh?`

lambdas/backend-api/package.json
```diff
@@ -3,23 +3,26 @@
   "private": true,
   "version": "0.0.1",
   "scripts": {
-    "test:unit": "jest",
+    "test:unit": "NODE_OPTIONS=\"$NODE_OPTIONS --experimental-vm-modules\" npx jest",
```
alexnuttall, line 6: we already have a `jest.setup.ts` in this package, can that be used?

lambdas/backend-api/src/templates/infra/template-repository.ts
```diff
@@ -234,11 +240,82 @@ export class TemplateRepository {
...
+  async setLetterValidationResult(
+    templateKey: TemplateKey,
+    versionId: string,
+    valid: boolean,
+    personalisationParameters: string[],
+    csvHeaders: string[]
+  ) {
+    const ExpressionAttributeNames: UpdateCommandInput['ExpressionAttributeNames'] =
+      {
+        '#files': 'files',
```
chris-elliott-nhsd, line 252: you don't need ExpressionAttributeNames unless the attribute name is a reserved word (ie status), most of these are unnecessary

lambdas/backend-api/src/templates/domain/template-pdf.ts
```diff
@@ -0,0 +1,105 @@
...
+        const textContent = await page.getTextContent();
+
+        // page.getTextContent call above can only return TextItem objects (includeMarkedContent option defaults to false)
+        // But method signature returns (TextItem | TextMarkedContent)[]
+        // Have to use type-guard to get the right typing
+        const parameters = textContent.items
+          .filter((item) => isTextItem(item))
+          .map((item) => item.str)
+          .join('')
+          .matchAll(/\(\([^(]*?\)\)/g);
```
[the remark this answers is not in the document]
chris-elliott-nhsd: disregard, i've seen that we're separately validating that the personalisation is sensibly formatted

lambdas/backend-api/src/templates/domain/validate-letter-template-files.ts
```diff
@@ -0,0 +1,50 @@
...
+    pdf.addressLinePersonalisationParameters.length ===
+    ADDRESS_PERSONALISATIONS.length;
+
+  const correctAddressLines = pdf.addressLinePersonalisationParameters.every(
+    (line, i) => line === ADDRESS_PERSONALISATIONS[i]
+  );
+
+  const customParametersSensiblyFormatted =
+    pdf.customPersonalisationParameters.every((parameter) =>
+      /^\w+$/.test(parameter)
```
chris-elliott-nhsd, line 18: i can't find it documented anywhere that this is precisely what we mean by 'sensibly formatted'. do we need to update the official documentation?

lambdas/backend-api/src/templates/domain/validate-letter-template-files.ts
```diff
@@ -0,0 +1,50 @@
...
+    );
+
+  const requiredTestFileExists =
+    pdf.customPersonalisationParameters.length === 0 || Boolean(csv);
+
+  let testFileHasCorrectNumberOfColumns = requiredTestFileExists;
```
chris-elliott-nhsd, line 24: nitpicky but i feel like you don't need `let` and reassigning here. just return here if you don't have a test data CSV

infrastructure/terraform/modules/backend-api/cloudwatch_event_rule_guardduty_quarantine_scan_no_threats.tf
harrim91: This is quite a long comment that's turned into a bit of a rubber-duck. Thanks 😁 

Simpler lambdas is a big motivation. But it's mainly around the ordering that we want to do things. 

When the scan happens, there are 4 things we want to happen - 4 lambdas.

Scan passes - set the file status, copy the file to internal, validate the files

Scan fails - set the file status, delete the file

I think without the ordering requirements, that could all be done just off two rules from the GuardDuty event. One for passed, one for failed. 

In the passing scenario, the validation has dependencies on the copy and the set-status. So the validation gets triggered by an event which is emitted in the set-status lambda. So we need an extra rule for that event. 

In the failure scenario, set status actually reads the file with HeadObject to get metadata from it. If the delete lambda was also responding to the same GuardDuty event, there'd be a race condition and the file might be deleted before we could get its metadata. 

So there are failure scenarios and passing scenarios that we need to handle on both the GuardDuty event and on the custom event. 

I think with the failure scenario - we _could_  remove the `HeadObject` call parse the data from the file path - I think everything that the set-status lambda needs is in the path. In which case, the delete lambda could respond to the GuardDuty failure event too rather than the custom event. 

With the validation lambda. There are checks in the lambda to make sure that all the required database updates and file copies have run. So it _could_ run in response to the GuardDuty event too I suppose. You'd probably see more failures and retries in the Lambda, but it's designed to handle those anyway so maybe I can simplify it as you've suggested. I can try and re-jig it a bit and see what happens. It'd mean I don't have to emit the custom event at all, and remove a few more bits of infrastructure. I'm keen to simplify it as much as possible, so will give this a try. Thanks Chris!

infrastructure/terraform/modules/backend-api/locals.tf
```diff
@@ -15,14 +17,30 @@ locals {
...
+
+  backend_lambda_environment_variables = {
+    ENABLE_LETTERS_BACKEND           = var.enable_letters
+    ENVIRONMENT                      = var.environment
+    NODE_OPTIONS                     = "--enable-source-maps"
+    TEMPLATES_QUARANTINE_BUCKET_NAME = module.s3bucket_quarantine.id
+    TEMPLATES_INTERNAL_BUCKET_NAME   = module.s3bucket_internal.id
+    TEMPLATES_EVENT_BUS_NAME         = data.aws_cloudwatch_event_bus.default.name
+    TEMPLATES_EVENT_SOURCE           = local.event_source
+    TEMPLATES_TABLE_NAME             = aws_dynamodb_table.templates.name
```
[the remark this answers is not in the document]
harrim91: The lambdas all get initialised with the `createHandler(createContainer())` pattern. 

Because they are all sharing a container, each lambda gets a `TemplateRepository`, a `LetterUploadRepository` a `TemplateClient`, an `EventClient` instance created on initialization, even if it doesn't use them. And each of those things has config that it requires. 

e.g. Everything now has access to an `EventClient` instance, even if it doesn't use it. I added validation that the required config was set. But that was causing all of the other lambdas that didn't actually use the EventClient to fail.

There was some validation in `createContainer` already which was checking that required env vars were present. It was a bit untidy and getting more complicated when adding more variables. So I created a single zod validator that expects all of the lambdas to have the same set of env vars. It's maybe a bit odd, but it's simpler and it's quite nice, as the multiple lambdas are all effectively forming one application, and it means that if you look at the bigger picture, then the whole application now has the same configuration. 

There are also env vars that it's good to have in all the lambda I think - they all now have `NODE_OPTIONS`, they all have the letters feature flag, and the environment name etc. 

An alternative would be to have a more streamlined container per lambda. I wouldn't be against that necessarily, but it feels like a separate change maybe

infrastructure/terraform/modules/lambda-function/variables.tf
harrim91: This is just based on the default that lambda gives you if you don't set it. This module didn't expose the timeout parameter before, so I'm guessing that all of our Lambda's created with this module will currently have a 3 second timeout

lambdas/backend-api/package.json
harrim91: I'll give it a go, but I don't think it'll work. This is setting the `NODE_OPTIONS` which is used by the Node process which Jest runs in - it's not a config option for Jest itself.

lambdas/backend-api/src/templates/domain/validate-letter-template-files.ts
harrim91: "Sensibly formatted" is a vague requirement isn't it 😆 I'll raise it and see if we can get the docs amended. Thanks.

lambdas/backend-api/src/templates/domain/validate-letter-template-files.ts
harrim91: Ah, there was a couple of reasons for this.

One - I think if you have a PDF with no personalisation, but you still upload a test file with data, we wanted to fail.

So the conditions in the `if (csv)` block will catch that scenario (`testFileHasCorrectNumberOfColumns` will be `false`). 


Also because I wanted to add a log like:

```
{
    correctAddressLength,
    correctAddressLines,
    customParametersSensiblyFormatted,
    requiredTestFileExists,
    allCustomPersonalisationIsInTestFile,
    testFileHasCorrectNumberOfColumns,
    allTestFileRowsContainCorrectNumberOfColumns,
}
```

But I clearly forgot to do that. 

I'll have a think if I can write it a bit more expressively.

layers/pdfjs/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "nhs-notify-web-template-management-lambda-layer-pdfjs",
+  "private": "true",
+  "version": "5.1.91",
+  "main": "index.ts",
+  "dependencies": {
+    "pdfjs-dist": "5.1.91",
+    "@napi-rs/canvas": "0.1.67",
+    "@napi-rs/canvas-linux-x64-gnu": "0.1.67"
```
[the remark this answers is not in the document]
harrim91: The simple answer is that it doesn't work. It took me many many hours to get this stupid `pdfjs-dist` library working.

There are multiple complicating factors. 

For why it has to be a layer - we're using esbuild, which bundles and tree-shakes, which conflicted with something about how pdfjs-dist loads its dependencies internally, meaning it was still looking in the nonexistent `node_modules` rather than in the bundled code. 

So I had to mark `pdfjs-dist` as external from the main bundle, so esbuild doesn't bundle it, and it still gets loaded from `node_modules`

And it also has a dependency on `@napi-rs/canvas` which needs a native binary which is os/architecture specific. So I had to put the specific version which works with the Lambda runtime so we can get the right version downloaded regardless of the platform that is running the build. 

There was also loads of pain around `pdfjs-dist` only exporting esm rather than cjs. But I don't think that bit necessitated the layer. Just an additional pain and tweaks to ts, jest and esbuild configs.

lambdas/backend-api/src/templates/api/set-letter-file-virus-scan-status.ts
```diff
@@ -4,15 +4,16 @@ import {
...
-      metadata: Omit<
-        LetterUploadMetadata,
-        'user-filename' | 'test-data-provided'
-      >;
+      bucketName: string;
+      objectKey: string;
+      versionId: string;
     };
     scanResultDetails: {
       scanResultStatus: GuardDutyMalwareScanStatus;
```
chris-elliott-nhsd, line 19: late addition to the code review but i just noticed it while doing some work that builds on this branch. `GuardDutyMalwareScanStatus` is derived from a type in a package where the status is called `THREAT_FOUND` but our code is looking for a status called `THREATS_FOUND`

lambdas/backend-api/package.json
harrim91: No I think it needs to be like this 😞

tests/test-team/fixtures/pdf-upload/with-personalisation/test-data.csv
harrim91: Done. I've done it in line with the sheet available for download on that link, which has these headers:

```csv
Personalisation field,Short length data example,Medium length data example,Long length data example
```

lambdas/backend-api/src/templates/infra/template-repository.ts
harrim91: We're using these in most of our existing DynamoDB interactions already and I kind of like the consistency of using them. It means you don't have to worry about tripping over reserved words.

lambdas/backend-api/src/templates/domain/validate-letter-template-files.ts
harrim91: I can't really think of a way to do this that I prefer

infrastructure/terraform/modules/lambda-function/variables.tf
```diff
@@ -53,8 +53,33 @@ variable "memory_size" {
...
+  default     = null
+}
+
+variable "sqs_event_source_mapping" {
+  description = "Configuration for SQS event source mapping"
+  type = object({
+    sqs_queue_arn                      = string
+    batch_size                         = optional(number, 10)
+    maximum_batching_window_in_seconds = optional(number, 0)
+    scaling_config = optional(object({
```
alexnuttall, line 80: Didn't we make this required in main? Do we need similar here?

lambdas/backend-api/src/templates/api/validate-letter-template-files.ts
```diff
@@ -0,0 +1,208 @@
...
+      'version-id': versionId,
+    } = metadata;
+
+    const getTemplateResult = await this.templateRepository.get(
+      templateId,
+      owner
+    );
+
+    if (getTemplateResult.error) {
+      log.error('Unable to load template data', getTemplateResult.error);
```
alexnuttall, line 92: Does the error logging work here?

chris-elliott-nhsd (approved):

tests/test-team/template-mgmt-e2e-tests/template-mgmt-letter-file-validation.e2e.spec.ts
```diff
@@ -87,6 +110,71 @@ test.describe('letter file validation', () => {
     }).toPass({ timeout: 100_000 });
```
alexnuttall, line 110: Not your change, but this looks wrong

alexnuttall (dismissed):

lambdas/backend-api/src/templates/api/validate-letter-template-files.ts
harrim91: Yep. the `getTemplateResult.error` isn't an actual error instance, it's a `FailureResult` thing from `lambdas/backend-client/src/types/result.ts`. I will send you a link to an example of this log in CloudWatch.

ClareJonesBJSS (approved):

alexnuttall:
@harrim91 I'm not suggesting a change post-testing - but the field name `csvHeaders` could be a bit misleading. The actual headers file in core would be a comma-joined version of `personalisationParameters`.

ClareJonesBJSS (dismissed):

alexnuttall (dismissed):

gparry333 (dismissed):

infrastructure/terraform/modules/lambda-layer/null_resource_package_layer.tf
aidenvaines-bjss, line 1: This is making me feel uneasy. Can this be moved up into that pre-sh file please. In a future change we will want to break out the lambda packaging mechanisms and layers being done like this would present a number of challenges to back-out of.

infrastructure/terraform/modules/lambda-layer/README.md
aidenvaines-bjss, line 1: See comments on the null_resource, but this whole module makes me uneasy.

Pack layer for layers in /mylayers or whatever in your pre-sh then use a lambda_layer_version in the backend-api module to handle it would be my preference

infrastructure/terraform/modules/backend-api/cloudwatch_event_rule_guardduty_quarantine_scan_failed.tf
```diff
@@ -0,0 +1,56 @@
...
+  role_arn = aws_iam_role.quarantine_scan_failed.arn
+}
+
+resource "aws_cloudwatch_event_target" "quarantine_scan_failed_delete_object" {
+  rule     = aws_cloudwatch_event_rule.quarantine_guardduty_scan_failed.name
+  arn      = module.lambda_delete_failed_scanned_object.function_arn
+  role_arn = aws_iam_role.quarantine_scan_failed.arn
+}
+
+resource "aws_iam_role" "quarantine_scan_failed" {
```
aidenvaines-bjss, line 33: Separate out your role into its own resource please
Amazon recommend the usage of Managed Policies over Inline policies https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_managed-vs-inline.html 

>  In most cases, we recommend that you use managed policies instead of inline policies.

This also got picked up on the pen-test and i spent 2 days remediating the inline policies

infrastructure/terraform/modules/backend-api/cloudwatch_event_rule_guardduty_quarantine_scan_passed.tf
```diff
@@ -0,0 +1,79 @@
...
+  role_arn = aws_iam_role.guardduty_quarantine_scan_passed.arn
+}
+
+resource "aws_cloudwatch_event_target" "guardduty_quarantine_scan_passed_validate_files" {
+  rule     = aws_cloudwatch_event_rule.guardduty_quarantine_scan_passed.name
+  arn      = module.sqs_validate_letter_template_files.sqs_queue_arn
+  role_arn = aws_iam_role.guardduty_quarantine_scan_passed.arn
+}
+
+resource "aws_iam_role" "guardduty_quarantine_scan_passed" {
```
aidenvaines-bjss, line 39: image

infrastructure/terraform/modules/backend-api/locals.tf
```diff
@@ -1,7 +1,9 @@
 locals {
   csi = "${var.csi}-${var.module}"
 
-  lambdas_source_code_dir = abspath("${path.module}/../../../../lambdas")
+  monorepo_root           = abspath("${path.module}/../../../..")
```
aidenvaines-bjss, line 4: Monorepo? naming revision please repo_root is sufficient
```suggestion
  repo_root           = abspath("${path.module}/../../../..")
```

aidenvaines-bjss, line 4: Monorepo? naming revision please repo_root is sufficient
```suggestion
  repo_root           = abspath("${path.module}/../../../..")
```

infrastructure/terraform/modules/backend-api/module_build_template_lambda.tf
```diff
@@ -11,7 +11,12 @@ module "build_template_lambda" {
     local.backend_lambda_entrypoints.delete_template,
     local.backend_lambda_entrypoints.list_template,
     local.backend_lambda_entrypoints.set_file_virus_scan_status,
+    local.backend_lambda_entrypoints.copy_scanned_object_to_internal,
```
aidenvaines-bjss, line 14: lets alphabetise this list please

infrastructure/terraform/modules/backend-api/locals.tf
```diff
@@ -17,15 +19,27 @@ locals {
...
+    get_template                    = "src/templates/get.ts"
+    list_template                   = "src/templates/list.ts"
+    set_file_virus_scan_status      = "src/templates/set-letter-file-virus-scan-status.ts"
+    submit_template                 = "src/templates/submit.ts"
+    template_client                 = "src/index.ts"
+    update_template                 = "src/templates/update.ts"
+    validate_letter_template_files  = "src/templates/validate-letter-template-files.ts"
+  }
+
+  backend_lambda_environment_variables = {
```
aidenvaines-bjss, line 36: Good shout

aidenvaines-bjss (changes requested):
Platform review of infra../terraform

lambdas/layers/pdfjs/build.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
aidenvaines-bjss, line 1: Lovely work Clare

aidenvaines-bjss (approved):

chris-elliott-nhsd (approved):

PR #454, opened 2025-05-06
CCM-9749: Client side validation

bhansell1:
## Description

- Moves form error messages into `content.ts`
- Adds client-side form validation (we still do server-side validation)

## Context

We are only doing form validation on the server which can be slow (relatively) feedback for a user. Doing validation on the client-side results in faster (relatively) feedback for the user. 

## Type of changes

- [x] Refactoring (non-breaking change)

## Checklist

---


frontend/src/content/content.ts
```diff
@@ -509,6 +592,19 @@ const templateFormSms = {
     'Learn more about character counts and text messaging pricing (opens in a new tab)',
   buttonText: 'Save and preview',
   backLinkText: 'Back to choose a template type',
+  form: {
+    smsTemplateName: {
+      error: {
+        empty: 'Enter a template name',
```
alexnuttall, line 598: Should these pieces of shared text have a single source e.g. `const enterTemplateName` if we don't do that, I struggle to see the value in this file

alexnuttall (approved):

ClareJonesBJSS (dismissed):

chris-elliott-nhsd (dismissed):

frontend/src/content/content.ts
bhansell1: Good shout! Updated.

alexnuttall (approved):

frontend/src/content/content.ts
ClareJonesBJSS: For what its worth, I included moving all the messages in here in the ticket as a general design principle. While the biggest advantage ime for having all displayed text separated out (easier to convert into JSON compatible with an i18n lib) seems unlikely to come up for us, it can still be really helpful to try and guarantee that every word the end user sees can be found and thus proof-read/amended in one place, rather than scattered across a code base.

PR #457, opened 2025-05-06
CCM-9872: Copy file to download bucket

chris-elliott-nhsd:
## Description

PR to copy files to download bucket if there's a successful virus scan

## Type of changes

## Checklist

---

### DIFF
--- a/lambdas/backend-api/src/templates/api/process-proof.ts
+++ b/lambdas/backend-api/src/templates/api/process-proof.ts
@@ -36,23 +36,24 @@ export const createHandler =
       scanResultStatus === 'NO_THREATS_FOUND' ? 'PASSED' : 'FAILED';
 
     if (virusScanResult === 'PASSED') {
-      await letterFileRepository.copyFromQuarantineToInternal(
-        objectKey,
-        versionId,
-        internalKey
-      );
-
       const downloadKey = LetterProofRepository.getDownloadKey(
         owner,
         templateId,
         fileName
       );
 
-      await letterFileRepository.copyFromQuarantineToDownload(
-        objectKey,
-        versionId,
-        downloadKey
-      );
+      await Promise.all([
+        letterFileRepository.copyFromQuarantineToInternal(
+          objectKey,
+          versionId,
+          internalKey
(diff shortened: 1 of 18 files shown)


lambdas/backend-api/src/templates/api/process-proof.ts
```diff
@@ -41,6 +41,18 @@ export const createHandler =
...
         internalKey
       );
+
+      const downloadKey = LetterProofRepository.getDownloadKey(
+        owner,
+        templateId,
+        fileName
+      );
+
+      await letterFileRepository.copyFromQuarantineToDownload(
```
bhansell1, line 51: The ticket mentions the proof is should be copied form internal to download, has that changed somewhat?

chris-elliott-nhsd: i didn't think too much about it as i don't think it makes any difference where the copy comes from as it is happening at the same time as the copy to internal. i guess having quarantine as the copy source means you could do the two copies in parallel and save a bit of time (which i haven't done but might now you've made me think of it), do you think there are good reasons to copy from internal?

bhansell1: I guess I'm a little out-of-touch with the process.

My initial reaction is that it seems wrong to copy a file from a bucket called quarantine to download. It doesn't extend further than that. But I don't see any issues with it.

We have the `copy-scanned-object-to-internal.ts` which from looking at it would copy the proofs to internal. I don't understand why `process-proof` needs to copy to internal. Most likely I'm missing something.

bhansell1: Jumped on a call.

Thanks for explaining the reason why `copy-scanned-object-to-internal.ts` is not used here.

bhansell1 (approved):

alexnuttall (approved):

aidenvaines-bjss (dismissed):

alexnuttall (approved):

aidenvaines-bjss (approved):

PR #529, opened 2025-06-23
CCM-10333: save clientId on templates

alexnuttall:
## Description

In the APIG authorizer, `clientId` is extracted from the access token if it's present, and forwarded to the backend. `clientId` is added to the template on creation.

The ticket says to add `userId` to the template, but it's unclear if this should be the creator or most recent updater - so I have added both `updatedBy` and `createdBy`. Both of these are optional fields - though we could populate them as part of the `owner` data migration, and then make them required.

Instead of requiring sandbox users to be added to groups, I have added a custom attribute to the sandbox user pool: `sbx:client_id` which can be set on users at creation, or afterwards. The templates sandbox has it's own pre-token-generation lambda which reads the custom attribute and sets `nhs-notify:client-id` in the access token.

## Type of changes

## Checklist

---


lambdas/backend-api/src/templates/app/template-client.ts
```diff
@@ -32,9 +31,10 @@ export class TemplateClient implements ITemplateClient {
 
   async createTemplate(
     template: CreateUpdateTemplate,
-    owner: string
+    userId: string,
```
harrim91, line 34: I wonder if it'd be cleaner to pass the user id and client id around together as an object - call it `authContext` or something. It might make it a bit clearer where those values are coming from in the lower level code and in the logs.

harrim91 (dismissed):
This is great. One suggestion, but happy as it is.

lambdas/backend-api/src/templates/app/template-client.ts
```diff
@@ -375,18 +394,17 @@ export class TemplateClient implements ITemplateClient {
   private async updateTemplateStatus(
     templateId: string,
     status: Exclude<TemplateStatus, 'SUBMITTED' | 'DELETED'>,
-    owner: string
+    userId: string,
+    logger: Logger
```
chris-elliott-nhsd, line 398: why does this want its own logger when there's a class logger?

lambdas/backend-api/src/templates/app/template-client.ts
```diff
@@ -21,7 +20,7 @@ import { Logger } from 'nhs-notify-web-template-management-utils/logger';
 import { LetterUploadRepository } from '../infra/letter-upload-repository';
 import { ProofingQueue } from '../infra/proofing-queue';
 
-export class TemplateClient implements ITemplateClient {
```
chris-elliott-nhsd, line 24: :tada:

lambdas/cognito-triggers/src/pre-token-generation.ts
```diff
@@ -0,0 +1,67 @@
...
+    const clientId = event.request.userAttributes['custom:sbx:client_id'];
+
+    if (clientId) {
+      response = PreTokenGenerationLambda.setTokenClaims(event, 'accessToken', {
+        'nhs-notify:client-id': clientId,
+      });
+
+      response = PreTokenGenerationLambda.setTokenClaims(event, 'idToken', {
+        'nhs-notify:client-id': clientId,
+      });
```
chris-elliott-nhsd, line 33: why does it need to be on both tokens?

alexnuttall: I'm just matching what was done in IAM: https://github.com/NHSDigital/nhs-notify-iam-webauth/commit/2e676afd1cf29404a49b8e651bc8d19a59e801b4

lambdas/backend-api/src/templates/app/template-client.ts
alexnuttall: removed

lambdas/backend-api/src/templates/app/template-client.ts
alexnuttall: added a `User` object here

chris-elliott-nhsd (dismissed):

infrastructure/terraform/modules/cognito-triggers/module_pre_token_generation_lambda.tf
```diff
@@ -0,0 +1,37 @@
...
+  project        = var.project
+  environment    = var.environment
+  component      = var.component
+  aws_account_id = var.aws_account_id
+  region         = var.region
+
+  function_name = "pre-token-generation"
+  description   = "Pre token generation handler for Cognito user pool"
+
+  # function code
```
aidenvaines-bjss, line 13: exciting use of multiple comment prefixes and comment string. are these strictly necessary?

`/*` & `*/` also work for multilines too if you want the other one :D

aidenvaines-bjss (changes requested):

infrastructure/terraform/modules/cognito-triggers/module_pre_token_generation_lambda.tf
alexnuttall: removed comments

PR #624, opened 2025-08-11
CCM-11027: enforce correct versioning of event schemas

harrim91:
## Description

Adds CI checks to ensure that the event schema package has been versioned correctly when changes are made, and that schemas have been regenerated if required.

## Type of changes

## Checklist

---

### DIFF
--- a/scripts/is_valid_increment.sh
+++ b/scripts/is_valid_increment.sh
@@ -1,3 +1,6 @@
+#!/usr/bin/bash
+# Determines whether a version increment to a package is valid
+
 is_valid_increment() {
   local current="$1"
   local next="$2"


alexnuttall (dismissed):

ClareJonesBJSS (dismissed):

.github/workflows/stage-1-commit.yaml
```diff
@@ -199,3 +199,82 @@ jobs:
...
+      - name: "Checkout code"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Detect package changes
+        id: check
+        run: |
+          git fetch origin main
+          if git diff --quiet origin/main...HEAD -- packages/event-schemas; then
```
chris-elliott-nhsd, line 221: we're comparing against git here but against the published package later on - this feels like it would not be robust in the event of eg pipeline failures that could cause git and the published package to be out of sync. why not compare to the published package everywhere?

chris-elliott-nhsd: i didn't have a specific idea in mind, i just wasn't sure about using different sources for determining whether a change happened and whether the change had a valid version. now i'm thinking about it, comparing to git main in both places might make more sense than comparing to the published package given that's what we want to use as our single source of truth?

as we're looking at the main branch here anyway, can we get the version used in main here and pass that to is_valid_increment?

also this isn't very important so feel free to just ignore me if you've got better things to do

harrim91: Good shout, I like the idea of using Git for everything at this point. I will make the change.

harrim91: Updated

chris-elliott-nhsd (dismissed):

PR #680, opened 2025-09-10
CCM-10981: cache node_modules

bhansell1:
## Description

- caches node_modules using  a hash of `package.lock.json` as the cache key. 
- restores cache if the cache key is hit 
- splits `Test Stage / Unit-tests` into different runners per workspace to run in parallel at the same time

## Context

Doing `npm ci` everywhere was causing the pipelines to spend on average **1 minute 30 seconds** to install. To speed up the pipeline the decision was to cache the `node_modules`. 

`make test-unit` was taking up **4 minutes** to finish executing. This was because `npm workspaces` are run in series, there are no options or parallelism. Investigated use of `pnpm workspaces` which does support `parallelism` but no real improvement was made for the headache of switching packing managers ([link to branch](https://github.com/NHSDigital/nhs-notify-web-template-management/compare/main...feature/CCM-10981_pnpm)). 

Using GitHub's matrix strategy offered up a nice solution to parallelism. Using `package.json` workspaces we could run the unit tests for each workspace with no additional maintenance. Going for this approach has shaved about **2 minutes** from CI. 

I haven't applied the matrix strategy to Linting or Type checking since both steps are already fast enough. 

https://nhsd-jira.digital.nhs.uk/browse/CCM-12191  - raised for further improvements

## Type of changes

## Checklist

---

### DIFF
--- a/scripts/tests/unit.sh
+++ b/scripts/tests/unit.sh
@@ -4,7 +4,6 @@ set -euo pipefail
 
 cd "$(git rev-parse --show-toplevel)"
 
-# read optional first arg; empty means "run all workspaces"
 WORKSPACE="${WORKSPACE:-}"
 
 # This file is for you! Edit it to call your unit test suite. Note that the same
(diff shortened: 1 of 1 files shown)


bhansell1:
> Approving this - although wondering if it wold be worth seeing if the caching can be plugged into the jobs in internal repo too? The terraform sandbox job and the testing job. It'd definitely be viable in the testing job. The terraform jobs do the install in `pre.sh` rather than in the workflow directly. So that might be more tricky. There's the `$SKIP_SANDBOX_INSTALL` option though. You could install from the cache and then run the Terraform job with that set to true for these sandbox jobs.

Yeah I raised [CCM-12191](https://nhsd-jira.digital.nhs.uk/browse/CCM-12191) to address those similar concerns.

chris-elliott-nhsd (approved):

